PR #16, opened 2020-11-27
location_access: Reduce overall size of pairing classes

lioncash:
While we're at it, we can place an interface around them to make for slightly nicer reading.


erinexplosives:
I agree

PR #30, opened 2020-12-24
Logic:Added Beginnings of MQ Logic and Feature to Exclude Locations

gymnast86:
- Defined all MQ locations and implemented Deku Tree MQ logic
- New Feature to exclude Item Locations from having advancement items (looks bad visually, but works)
- Finally finished option to start at either day or night


source/item_pool.cpp
```diff
@@ -484,6 +483,19 @@ static Item GetJunkItem() {
   return JunkPoolItems[idx];
 }
 
+int PlaceJunkInExcludedLocation(ItemLocation * il) {
```
lioncash: The return value of this doesn't seem to be used anywhere, was that intended (e.g. for use later on), or should it be void?

gymnast86: It should just be void. I probably intended to use the return value for something while programming but didn't end up using it and then forgot about it.

PR #45, opened 2021-03-20
Changed reachability search

cjohnson57:
Now items found during the search are cached and added before the next search iteration. This prevents needing multiple tries of iterations with no new locations found.

I generated 10 seeds with this change and as far as I can tell they all looked correct.


gymnast86 (approved):
This looks good and should help cut down the time it takes to place items on console.

PR #48, opened 2021-03-20
Reworked RNG to use Mersenne Twister

cjohnson57:
rand() is not considered a good way to produce randomness, mainly due to distribution of numbers not being even and the PRNG algorithm behind it not being great.

I changed the RNG to use Mersenne Twister, which is a standard PRNG in many things such as Python. I also made some helpful functions utilizing the generator, for getting a random element from and shuffling vectors/arrays.

Finally, I further reworked the reachability search a little bit because I still wasn't quite happy with it.


cjohnson57:
Turns out my previous PR breaks key logic but my further changes to the reachability search in this PR (should) fix this

gamestabled (approved):

PR #52, opened 2021-03-21
App: Include new build/release tool using actions.

PhlexPlexico:
Update custom_messages.py, due to encoding issues on docker container.

This will create a new release on each push to the main branch. That way if people wish to beta test, they can use the latest code that is supposedly the most stable to test.


PhlexPlexico:
Just awaiting the png/wav files in order to generate the CIA. I got the RSF from the CIA file that was in releases and it should be using the same title id now (0x0008ad5b).

PhlexPlexico:
There, new banner audio and image are placed in. This should be good to go now.

PR #55, opened 2021-03-22
App: Update Actions To Include RomFS

PhlexPlexico:
This build includes making the RomFS using 3dstool (grabbed from its latest release on github). Verison number has been included in the build process, but a good TODO would be to set a Github secret or something-or-other to include in the action so we can set the build number without having to update now two places:

https://github.com/gamestabled/OoT3D_Randomizer/blob/be8781c505ff26f9c01268c76bb4a643e66ab622/.github/workflows/create_nightly.yml#L39

and

https://github.com/gamestabled/OoT3D_Randomizer/blob/be8781c505ff26f9c01268c76bb4a643e66ab622/source/menu.cpp#L37


PhlexPlexico:
I've made some changes in a different branch here, and debating moving them in. I've broken out the actual script into the repo so that people on Linux would be able to build with ease (apt installs all the needed libs, provided they have devkitarm), and creates an entire release for them. This allows us to easily edit the release number on the CIA as well, and includes a git SHA in the `menu.cpp` so that users can report which version they are using direct from the patcher. 

I've also updated to mark the dev builds as "pre-release" which should let the users know that they are not downloading an actual release build. If these changes are okay, I can merge it into the PR as well.

PR #57, opened 2021-03-22
App: Change to playthrough output and adding retries

cjohnson57:
1- Changed the playthrough output to display by spheres, and added all items which are advancement (with a few exclusions listed below). I believe this makes it easier to follow the logic of the seed, both for players and devs trying to make sure their changes didn't break anything (cough cough me). Here is an example of what a playthrough looks like now: https://pastebin.com/rPT5BD1H

The following are excluded:
1. Any shop "Buy" item
2. Gold skulltula tokens after the 50th
3. Bombchus after the first
4. Any items found in the same sphere or spheres after Ganon

The playthrough is still bloated by items not actually necessary to reach Ganon, as I can't currently find a way to check which items are strictly necessary efficiently.

2- Using the previous changes to playthrough which included adding a playthroughBeatable bool set to true when the triforce is found, added a retry feature when it's detected the playthrough is not beatable after calculating it. I used 5 retries from OoTR. (Assuming an actually just bad generation, this should be more than enough. TestRunner once told me the chance of a gen failing on OoT's world with Assumed Fill was ~1/100k though I'm not sure if that was a guesstimate or an actually tested value). Here's what it looks like when it fails 5 times in a row (the 5 counts up from 2): https://i.imgur.com/qVXtYDc.png

I tested that it would be able to successfully generate if it failed only once, and it did. That being said, I do ask if @gymnast86 can please check my changes to Fill to make sure everything that should be reset between attempts is reset.


gamestabled (approved):

source/location_access.cpp
```diff
@@ -2707,4 +2707,30 @@ namespace Exits { //name, scene, hint, events, locations, exits
         }
     }
   }
+
+  //Reset exits and clear items from locations
+  void ResetAllLocations() {
+    for (Exit* exit : allExits) {
+      exit->ResetVariables();
+      for (size_t i = 0; i < exit->locations.size(); i++) {
```
lioncash: This and the following indexing should be able to be compressed to:

```suggestion
      for (auto& locPair : exit->locations) {
```

couldn't it?

source/menu.cpp
```diff
@@ -586,8 +586,14 @@ void GenerateRandomizer() {
 
 	int ret = Playthrough::Playthrough_Init(finalHash);
 	if (ret < 0) {
-		printf("Error %d with fill. Press Select to exit.\n", ret);
-		return;
+    if(ret == -1) { //Failed to generate after 5 tries
```
lioncash, line 589: This looks a little wonky indentation-wise, were tabs and spaces intermixed on these lines?

cjohnson57: Indeed they were, not sure how that happened but it looked fine on my IDE so I didn't notice- Thanks for pointing it out

source/location_access.cpp
cjohnson57: Indeed it can, ty for pointing it out

PR #58, opened 2021-03-22
App: Expand Build Tools

PhlexPlexico:
Changes include:
- `menu.cpp` - Line has been added in for the git commit hash for easier debugging with users.
- Build script has been externalized so people pulling the project can build from source (provided they have devkitARM)
- RomFS is now included so the crashes on CIA should be resolved.


PhlexPlexico:
Good lord this was a messy fix for the build process. Takes longer, compiles python from source, simply because python 3.5.2 can't compile the messages for some reason. Maybe something to add to low priority to get an older version working?

PR #102, opened 2021-04-08
Logic: Add option for nighttime skulltulas expect sun's song

cjohnson57:
This adds an option to make it so night-exclusive gold skulltula locations will expect the player to have sun's song in logic. This is simply done by ANDing the following clause with every logic statement for a nighttime GS:
`(CanPlay(SunsSong) || !NightGSExpectSuns)`


gamestabled:
Neat addition that some players will appreciate. I feel like it would it be better to just make another logic variable called something like `CanGetNightTimeGS = (CanPlay(SunsSong) || !NightGSExpectSuns)` and use that in all of these places instead, to reduce all the repetitiveness

cjohnson57:
Definitely cleaner, good suggestion

PR #103, opened 2021-04-08
Logic:Changed some settings for future medallion shuffle

gymnast86:
- Added a (currently unsettable) option that specifies Link's Pocket item as a dungeon reward for the time being. This setting gives him the dungeon reward on file creation instead of pushing an override like the other future settings probably will.
- Added a clarification message that gauntlet colors will currently only apply to Link's model and not GID models.

### DIFF
--- a/source/fill.cpp
+++ b/source/fill.cpp
@@ -410,7 +410,7 @@ static void RandomizeDungeonRewards() {
   AssumedFill(rewards, dungeonRewardLocations);
 
   int baseOffset = I_KokiriEmerald.GetItemID();
-  u32 bitMaskTable[9] = {
+  static constexpr std::array<u32, 9> bitMaskTable = {
     0x00040000,
     0x00080000,
     0x00100000,
--- a/source/fill.cpp
+++ b/source/fill.cpp
@@ -423,11 +423,12 @@ static void RandomizeDungeonRewards() {
   };
 
   for (size_t i = 0; i < dungeonRewardLocations.size(); i++) {
-    rDungeonRewardOverrides[i] = dungeonRewardLocations[i]->GetPlacedItem().GetItemID() - baseOffset;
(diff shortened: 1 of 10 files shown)


source/fill.cpp
```diff
@@ -410,8 +410,25 @@ static void RandomizeDungeonRewards() {
   AssumedFill(rewards, dungeonRewardLocations);
 
   int baseOffset = I_KokiriEmerald.GetItemID();
+  u32 bitMaskTable[9] = {
```
lioncash:
```suggestion
  static constexpr std::array<u32, 9> bitMaskTable = {
```

source/fill.cpp
```diff
@@ -410,8 +410,25 @@ static void RandomizeDungeonRewards() {
...
+    0x00000010,
+    0x00000020,
+  };
+
   for (size_t i = 0; i < dungeonRewardLocations.size(); i++) {
     rDungeonRewardOverrides[i] = dungeonRewardLocations[i]->GetPlacedItem().GetItemID() - baseOffset;
+
+    //set the player's dugeon reward on file creation instead of pushing it to them at the start
+    if (i == dungeonRewardLocations.size()-1 /*&& LinksPocket.Is(LINKSPOCKET_DUNGEON_REWARD)*/) {
+      LinksPocketRewardBitMask = bitMaskTable[dungeonRewardLocations[i]->GetPlacedItem().GetItemID() - baseOffset];
```
lioncash: Maybe extract this to a local like:

```cpp
const auto index = dungeonRewardLocations[i]->GetPlacedItem().GetItemID() - baseOffset;
```

and use it here and in the assignment above at the start of the loop, so the expression isn't duplicated

gymnast86: Added in the next commit

PR #124, opened 2021-04-17
App: Implement Shopsanity on app side

cjohnson57:
This PR adds (commented for now) Shopsanity settings. The method by which shop items are placed was redone, with two vectors: One is a vector of vectors of shop locations, and the other is a vector of shop items, by default the vanilla shop items in the same order as the shops of the former vector. PlaceShopItems() iterates through the former vector placing items from the latter.

When shopsanity is off, the ShopItems vector simply keeps its default value so items are placed in their vanilla locations. 

When shopsanity is on, the ShopItems vector is shuffled before placing items in shops. Shopsanity 0 leaves it there. After placement, shopsanity 1-4 replaces indices on the left side of the shop with a NoItem and a randomly generated price. Shopsanity random chooses a random number from 1-4 for each shop. NoItem in the item location means that the shop location will be considered during the fill algo.

Even with shopsanity off, this PR affects shop locations by making them considered properly in logic, most notably tunics and blue fire can now be considered in logic. To accomplish this, the ConditionsMet() function was modified to also check the CanBuy function, which makes sure (if the item is a shop location) the player's wallet is big enough to buy it and some other conditions are met (in particular, that the player is an adult for tunic and has a bottle for bottle items). 

A price property was added to ItemLocation, mainly so prices for non-shop items placed in shops can be checked.

I also modified it so that shop location prices are printed in the spoiler.

Speaking of prices, they're currently set using a range of [0, 300] in increments of 5. This matches OoTR, but OoTR uses a betavariant distribution which I believe makes prices tend to hover around 100 or so. Here, every price in the range is currently equally likely, meaning right now Shopsanity will be more expensive on average than OoTR.

For now, shop locations are always hidden from the exclude list because I'm not really sure how they should be handled with the different settings.

As far as I can tell from looking at spoilers everything is working properly on the app side, although the patch will need to be changed to override shop locations.


cjohnson57:
By the way: Shop items are shuffled such that Buy tunics will not appear in child-only shops. This is what OoTR does. It's accomplished by repeatedly shuffling shop items until this isn't the case.

If we want to be different and simply allow child to buy adult tunics with a hack, this while loop can be removed, simplifying the whole process. (This is also less confusing, imo, since if a non-Buy tunic happens to end up in a shop it can be bought by child either way.)

PR #134, opened 2021-04-25
Setting to skip dampe race, other minor changes

HylianFreddy:
A setting is added to allow skipping the second race with Dampe. If it's selected, the hookshot chest will be spawned already when the player gets there. Dampe's patch previously checked if the chest had been opened to decide if the second race should start, now it checks for the chest to be spawned.

Removing the automatic text boxes with the previous PR made Navi turn green near the spots with the texts. This PR fixes that.

Windmill/Dampe's grave is added to the list of areas for which the warp songs restriction is removed.

Following a request from Alyx, getting an arrow item doesn't equip it automatically unless the inventory is full (basically arrows now behave like all other items when they're added to the inventory).


code/src/entrance.c
```diff
@@ -23,6 +23,7 @@ void Scene_Init(void) {
     memcpy(&gSceneTable[13], gSettingsContext.ganonsCastleDungeonMode          == DUNGEONMODE_MQ ? &gMQDungeonSceneTable[13] : &gDungeonSceneTable[13], sizeof(Scene));
 
     gRestrictionFlags[72].flags2 = 0; // Allows warp songs in GTG
+    gRestrictionFlags[93].flags2 = 0; // Allows warp songs in Windmill / Dampe's grave
```
gamestabled: I do think it's best to leave this out for now to preserve the trick with dampe's race. Can you either comment it out for now or just remove it?

gamestabled (changes requested):

code/src/entrance.c
HylianFreddy: Done. I commented it so it can be reenabled in the future without looking for that "93" again

PR #139, opened 2021-05-02
Settings:Minor fixes/additions

gymnast86:
- Modified fishing so that the rewards can be obtained regardless of what size the caught fish is (even if it isn't a new record).
- Added Starting Inventory options of Start with Consumables and Start with Max Rupees.
- Implemented the Tycoon's Wallet with a rupee capacity of 999 rupees to eventually be used in shopsanity.


source/patch.cpp
```diff
@@ -420,6 +420,36 @@ bool WritePatch() {
...
+  if (!R_SUCCEEDED(res = FSFILE_Write(code, &bytesWritten, totalRW, buf, 2, FS_WRITE_FLUSH))) {
+    return false;
+  }
+  totalRW += 2;
+
+  // Write Tycoon Wallet Capacity to code
+  if (!R_SUCCEEDED(res = FSFILE_Write(code, &bytesWritten, totalRW, &tycoonWalletCapacity, sizeof(tycoonWalletCapacity), FS_WRITE_FLUSH))) {
+    return false;
+  }
+  totalRW += sizeof(tycoonWalletCapacity);
```
gamestabled: it should be possible to make this a patch in patches.s. if you ctrl+f ".word" in patches.s you should find some examples of how you can make a patch which is just a constant. since its only an s16 you could probably just include the 2 bytes before it so that you can write a full word at 0x0053CC08

code/src/patches.s
```diff
@@ -1125,6 +1125,22 @@ MidoCheckDekuTreeClearFive_patch:
...
+.section .patch_FishingSizeIgnoreChild
+.global FishingSizeIgnoreChild_patch
+FishingSizeIgnoreChild_patch:
+    nop
+
+.section .patch_FishingSizeIgnoreAdult
+.global FishingSizeIgnoreAdult_patch
+FishingSizeIgnoreAdult_patch:
+    nop
+
```
gamestabled, line 1143: im pretty meh on making it so any fish works. its fine but i think we should revisit this in the future

source/patch.cpp
gymnast86: Oh cool, that's a lot easier than doing it how I was trying.

code/src/patches.s
gymnast86: Yeah, it isn't great, but I think it's better to do this now and revisit it later at some point.

PR #150, opened 2021-05-16
Improve preset file structure

### LINKED ISSUE #148: Improve preset file structure
When I looked into the preset file, I noticed that it has some room for improvements, including readability to the human eye. So I want to make the following suggestions for them:

### 1. Make them valid XML files

Currently the settings files are not valid XML, because the initial declaration and the root node are missing. This should be a rather easy fix, but will definitely help with external tools (which may be developed in the future), because managing valid XML files is simpler than arbitrary ones (which may not be supported by other languages' libraries).

### 2. Simplify the XML structure by moving setting name to an attribute

The current structure looks a little overkill for simple key-value-pairs. Instead of using sub-nodes, I suggest making the setting name an attribute, and keep the setting value as node content. So you would get something like this:

```xml
<!-- Current structure -->
<setting>
    <settingName><![CDATA[Logic]]></settingName>
    <valueName><![CDATA[Glitchless]]></valueName>
</setting>

<!-- Suggested structure -->
<setting name="Logic"><![CDATA[Glitchless]]></setting>
```

TinyXML should handle special characters like apostrophes and quotes correctly, so that the names should never break the XML file. Spaces are should be kept as well (as I noticed some setting names have spaces in front of them), has to be tested upon implementation.

### 3. Remove CDATA tags

While I can understand that CDATA tags are used to keep the XML valid, they are not necessary in our case. The only "dangerous" characters are >, < and &, which TinyXML should escape as well to their entity representation. Removing them makes the file easier to read to the human eye (e.g. when checking them for debugging).

------

### Summary (tl;dr)

So my complete suggestion is to change from this:

```xml
<setting>
    <settingName><![CDATA[Logic]]></settingName>
    <valueName><![CDATA[Glitchless]]></valueName>
</setting>

<setting>
    <settingName><![CDATA[Forest]]></settingName>
    <valueName><![CDATA[Open]]></valueName>
</setting>

<setting>
    <settingName><![CDATA[Door of Time]]></settingName>
    <valueName><![CDATA[Open]]></valueName>
</setting>
```

to this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<settings>
    <setting name="Logic">Glitchless</setting>
    <setting name="Forest">Open</setting>
    <setting name="Door of Time">Open</setting>
</settings>
```

If you agree with these suggestions, I would like to try to implement the changes myself. I will message you on Discord regarding this issue ;)

BluePsyduck:
The structure of the preset files has been optimized, cutting their size in half. More details to the changes can be found in #148.
This closes #148.

Please check the changes carefully, I am not a C++ developer and might made obvious mistakes. Changes were tested on a 3DS system (saving and loading of custom presets, and the CACHED_SETTINGS.xml variant and its cosmetic counterpart).

### DIFF
--- a/source/preset.cpp
+++ b/source/preset.cpp
@@ -124,7 +124,13 @@ bool LoadPreset(std::string_view presetName, OptionCategory category) {
     return false;
   }
 
-  XMLNode* curNode = preset.FirstChild();
+  XMLElement* rootNode = preset.RootElement();
+  if (strcmp(rootNode->Name(), "settings") != 0) {
+      // We do not have our <settings> root node, so it may be the old structure. We don't support that one anymore.
+      return false;
+  }
+
+  XMLElement* curNode = rootNode->FirstChildElement("setting");
 
   for (MenuItem* menu : Settings::mainMenu) {
     if (menu->mode != OPTION_SUB_MENU) {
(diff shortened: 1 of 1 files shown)


source/preset.cpp
```diff
@@ -90,24 +96,13 @@ bool SavePreset(std::string_view presetName, OptionCategory category) {
...
-      settingText->SetCData(true);
-      valueText->SetCData(true);
-
-      //add elements to the document
-      settingName->InsertEndChild(settingText);
-      valueName->InsertEndChild(valueText);
-      newSetting->InsertEndChild(settingName);
-      newSetting->InsertEndChild(valueName);
-      preset.InsertEndChild(newSetting);
+      newSetting->SetAttribute("name", std::string(setting->GetName()).c_str());
```
herwinw, line 101: This looks like a potential dangling pointer, or a needlessly complex statement.

Simplified explanation: the statement `std::string(settings->GetName())` builds a new `std::string` object, that will be destroyed after this line. The `.c_str()` operation returns a memory location that will be freed once that object is destroyed. If `SetAttribute` makes a copy of the value, the call is safe but complex, if it uses a pointer it might dangle.

In both cases, this should fix it:
```c++
newSetting->SetAttribute("name", setting->GetName().c_str());
```
(I'm assuming `settings->GetName()` returns a `std::string` (most likely a const reference), based on line 102. If it returns a character pointer, the `.c_str()` part can be dropped. Using the wrong value will result in a compile error, so it's pretty easy to figure out what you need.

BluePsyduck: `setting->GetName()` returns a `std::string_view`, which seems to not have a `c_str()` method itself, hence the conversion, I assumed. (It was already there, I only compacted the code a little when I changed it to the new XML structure.)

I also looked into the SetAttribute() function of TinyXML, and they seem to always make a copy of the string, so freeing the memory after this line should not be a problem (from what I understood).

PR #165, opened 2021-05-24
Slight cleanup in spoiler_log

herwinw:
Mostly making things `const`

### DIFF
--- a/source/spoiler_log.cpp
+++ b/source/spoiler_log.cpp
@@ -73,33 +73,29 @@ const RandomizerHash& GetRandomizerHash() {
   return randomizerHash;
 }
 
-static void SpoilerLog_SaveLocation(std::string_view loc, std::string_view item) {
-  logtxt += loc;
+static inline void SpoilerLog_AddFormatted(std::string_view header, std::string_view value) {
+  logtxt += header;
   logtxt += ": ";
 
   // Formatting for spoiler log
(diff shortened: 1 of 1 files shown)


source/spoiler_log.cpp
```diff
@@ -73,38 +73,34 @@ const RandomizerHash& GetRandomizerHash() {
   return randomizerHash;
 }
 
-static void SpoilerLog_SaveLocation(std::string_view loc, std::string_view item) {
-  logtxt += loc;
+static inline void SpoilerLog_AddFormatted(std::string_view header, std::string_view value) {
```
lioncash, line 76: inline won't do anything here, as the function is static already, so GCC is already going to be aggressive about inlining it where possible.

source/spoiler_log.cpp
```diff
@@ -23,7 +23,7 @@ namespace {
   std::string logtxt;
   std::string placementtxt;
 
-  std::array<std::string_view, 32> hashIcons = {
+  const std::array<std::string_view, 32> hashIcons = {
```
lioncash:
```suggestion
  constexpr std::array<std::string_view, 32> hashIcons = {
```

source/spoiler_log.cpp
gamestabled: no harm in being explicit about it though afaik

PR #166, opened 2021-05-24
Feature: Randomize all settings for Open, World, Shuffle, and Dungeon settings

ssroffe:
This randomizes all settings in the Open, World, Shuffle, and Shuffle Dungeon Items settings. If "randomize all settings" is cached, it will also re-randomize the settings on init. World ignores MQ dungeons following the 64 randomizer, and Open ignores Logic rules to avoid unwanted unbeatable seeds.

### DIFF
--- a/source/settings.cpp
+++ b/source/settings.cpp
@@ -1128,7 +1128,7 @@ namespace Settings {
       RandomGanonsTrials.SetSelectedIndex(ON);
     }
     else {
-      for (u8 i=2; i < openOptions.size(); i++) {
+      for (size_t i = 2; i < openOptions.size(); i++) {
         openOptions[i]->Unhide();
       }
     }
--- a/source/settings.cpp
+++ b/source/settings.cpp
@@ -1114,6 +1114,9 @@ namespace Settings {
   // Randomizes all settings in a category if chosen
   // Hides all relevant options
   void RandomizeAllSettings() {
+    //Init Random_ interface for consistency in racing
+    Random_Init(seed);
+
     // Open Settings
     if (RandomizeOpen) {
       // Skip Logic and RandomizeOpen Options to ensure proper logic
@@ -1122,7 +1125,7 @@ namespace Settings {
         openOptions[i]->Hide();
 
         //randomize options
-        openOptions[i]->SetSelectedIndex(rand() % openOptions[i]->GetOptionCount());
+        openOptions[i]->SetSelectedIndex(Random(0,openOptions[i]->GetOptionCount()));
       }
       // Randomize Ganon Trials
       RandomGanonsTrials.SetSelectedIndex(ON);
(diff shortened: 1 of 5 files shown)


source/settings.cpp
```diff
@@ -1088,6 +1111,87 @@ namespace Settings {
     ResolveExcludedLocationConflicts();
   }
 
+  // Randomizes all settings in a category if chosen
+  // Hides all relevant options
+  void RandomizeAllSettings() {
+    // Open Settings
+    if (RandomizeOpen) {
+      // Skip Logic and RandomizeOpen Options to ensure proper logic
+      for (u8 i=2; i < openOptions.size(); i++) {
```
lioncash:
```suggestion
      for (size_t i = 2; i < openOptions.size(); i++) {
```

source/settings.cpp
```diff
@@ -1088,6 +1111,87 @@ namespace Settings {
...
+        openOptions[i]->Hide();
+
+        //randomize options
+        openOptions[i]->SetSelectedIndex(rand() % openOptions[i]->GetOptionCount());
+      }
+      // Randomize Ganon Trials
+      RandomGanonsTrials.SetSelectedIndex(ON);
+    }
+    else {
+      for (u8 i=2; i < openOptions.size(); i++) {
```
lioncash:
```suggestion
      for (size_t i = 2; i < openOptions.size(); i++) {
```

Ditto for other loops below

source/settings.cpp
```diff
@@ -1088,6 +1111,87 @@ namespace Settings {
...
+        openOptions[i]->Unhide();
+      }
+    }
+
+    // World Settings
+    if (RandomizeWorld) {
+      // Skip RandomizeWorld Option
+      for (u8 i=1; i < worldOptions.size(); i++) {
+        // skip MQ options
+        if ((i == 4) || (i==5)) {
```
lioncash:
```suggestion
        if (i == 4 || i == 5) {
```

source/settings.cpp
```diff
@@ -1088,6 +1111,87 @@ namespace Settings {
...
+          continue;
+        }
+        worldOptions[i]->Hide();
+        //randomize options
+        worldOptions[i]->SetSelectedIndex(rand() % worldOptions[i]->GetOptionCount());
```
lioncash: Just curious why rand() is used instead of the Random_ interface

ssroffe: I saw gym use rand() as opposed to Random in his recent changes on the tunic colors, so I followed that. I wasn't sure if Random messes with the seed or not. I can change to that if it's preferred.

gymnast86: rand() was used for random cosmetic options because we don't want cosmetics interfering with the randomization of item placements. If two people are racing and one person wants to use specific gauntlet colors but the other person wants random colors, they would get different item placements if the random gauntlet colors used the Random_ interface to determine those. 

In this case, I think we might have to seed the Random_ interface with what the current seed is before randomizing the settings. Normally the seeding is done by combining the seed and the settings together, but we can't do that if the settings haven't been determined yet. So I'd probably add Random_Init(seed) before this and then use the Random_ interface so that it'll be consistent for players who may want to race random settings. The Random_ interface will get seeded again later, but it should always be the same if the initial seed is the same.

ssroffe: That makes total sense. I initialized the Random_ interface at the beginning of the function and changed the rest as you recommend in the latest commit (91bd980).

PR #167, opened 2021-05-24
Convert custom_messages constant functions to constant variables

herwinw:
C++20 could make those constexpr, but for now changing them to constant
variables does remove a lot of function calls.

### DIFF
--- a/source/custom_messages.cpp
+++ b/source/custom_messages.cpp
@@ -182,118 +182,140 @@ constexpr std::array DungeonColors = {
         return { data, size };
     }
 
+    const std::string MESSAGE_END          {"\x7F\x00"s};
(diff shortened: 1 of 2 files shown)


lioncash:
Part of the reason the use of functions here is desirable is because it removes a bunch of static constructors. GCC also [inlines](https://godbolt.org/z/4qYbEEdMT) functions used and defined in the same translation unit (assuming they're not huge), so I don't think the function overhead is as large as it seems.

For the strings variables, I think it would be nicer to make them either `constexpr char` arrays or `constexpr string_view` instances.

gamestabled:
As @lioncash pointed out this actually doesnt improve code gen. No real suggestions to improve it

PR #168, opened 2021-05-27
Workaround for endless "Placing items.."

BluePsyduck:
Working around faulty initialization of placedItem, leading to an endless "Placing items..", because the generator thinks that there is already an item placed at the location.

This change only fixes the effect, not the actual cause that placedItem is not correctly initialized to NoItem. If anybody has a fix for the actual cause, this workaround can be removed.

### DIFF
--- a/source/item_location.hpp
+++ b/source/item_location.hpp
@@ -82,6 +82,9 @@ public:
     }
 
     const Item& GetPlacedItem() const {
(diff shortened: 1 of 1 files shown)


gamestabled:
Thanks for this, because of incoming code changes we won't need to merge this but nice job 👍

PR #183, opened 2021-06-04
Shopsanity: Finalizing ice trap names in shops, fixing price distributions

cjohnson57:
* Greatly expands the items that ice traps can appear as (grandpa boomerang would be proud)
* Fixes bug in ReplaceMaxItem causing it to not modify ItemPool, causing ice traps to appear with ice traps off and presumably other issues
* Created table of trick names with easy, medium, and hard names for each item an ice trap can appear as (Thanks Charade!)
* Fixed shop and scrub price distributions using a lookup table of probability values

With this we can consider shopsanity provisionally finalized until more bugs/issues/desired changes come up, so that's cool


source/shops.cpp
```diff
@@ -178,24 +184,47 @@ std::vector<ItemKey> GetMinVanillaShopItems(int total_replaced) {
...
-//Average price = 126
-//~45% chance of needing no wallet, ~25% chance of needing 1, ~30% chance of needing 2
+//This table contains a cumulative probability for each possible shop price based on
+// a beta distribution with alpha = 1.5, beta = 2, and the result of the distribution, a float in [0.0, 1.0),
+// being mutliplied by 60, casted to an integer, then multiplied by 5 to give a value in range [0, 295] in increments of 5.
+// Meaning the first value is the probability of 0, the next value is the probability of 0 plus the probability of 5, etc.
+//Probabilities generated using a python script with 1 billion trials, so should hopefully be pretty good
+//Average price ~126
+//~38% chance of needing no wallet, ~45% chance of needing 1, ~17% chance of needing 2
+std::array<double, 60> ShopPriceProbability= {
```
lioncash: This and `ScrubPriceProbability` can probably be made constexpr

cjohnson57: You're correct, ty; I tried making them const and it didn't compile but forgot constexpr exists

PR #189, opened 2021-06-06
A few fixes in the Spanish translation

Dani88alv:


gamestabled:
@ other people with the power to do so: dont merge this yet, it seems like there was an oopsy with git and the ClearMessages function was accidentally deleted

PR #191, opened 2021-06-06
Fixed custom tunic red flashing when damaged

M-1-RLG:


gamestabled:
After some testing this seems to work for adult tunic but not child. lets see if that can be fixed before we merge it

PR #207, opened 2021-06-11
Farore's Wind Anywhere and No Ammo Drops options

CharadeNut:


code/src/entrance.c
```diff
@@ -28,6 +28,14 @@ void Scene_Init(void) {
 
     gRestrictionFlags[72].flags3 = 0; // Allows farore's wind in GTG
     gRestrictionFlags[94].flags3 = 0; // Allows farore's wind in Ganon's Castle
+    
+    //Allow Farore's Wind anywhere if given by settings
+    if (gSettingsContext.faroresWindAnywhere){
+        for (int i = 0; i < 99; i++){
+			gRestrictionFlags[i].flags3 &= ~0x30;
```
lioncash: We use spaces instead of tabs for indentation

code/src/item00.c
```diff
@@ -16,7 +16,30 @@
 
 void EnItem00_rInit(Actor* thisx, GlobalContext* globalCtx) {
     EnItem00* item = THIS;
-
+	s16 item00Type = item->actor.params & 0x00FF;
```
lioncash: Ditto here, where applicable for relevant lines.

code/src/entrance.c
```diff
@@ -28,6 +28,14 @@ void Scene_Init(void) {
...
+			gRestrictionFlags[i].flags3 &= ~0x03;
```
HylianFreddy: This line makes it possible to use every item anywhere, not only FW. Is that intentional?

source/logic.cpp
```diff
@@ -430,6 +437,7 @@ namespace Logic {
     BlueFire     = HasBottle && BlueFireAccess;
     Fish         = HasBottle && FishAccess;
     Fairy        = HasBottle && FairyAccess;
+	AmmoDrops    = NoDrops.IsNot(NODROPS_NOAMMO) && NoDrops.IsNot(NODROPS_NOAMMOORHEALTH);
```
lioncash: Seems like tabs have snuck in here as well

gamestabled:
regardless of above code comments and questions (i know some discussion happened in discord as well) i think we should freeze this until after 2,0

PR #234, opened 2021-06-26
Fix startup crashes by splitting up individual patches if larger than 65535 bytes

gymnast86:
The IPS patching specification only allows individual patches to be 65535 bytes at most. This was causing startup crashes when an individual patch was larger. If patches are lager than 65535 bytes they will be split up into chunks so that the IPS patching can handle them properly.

- Also includes logic fix for Fire Temple in Entrance Randomizer


code/src/entrance.c
```diff
@@ -88,6 +88,13 @@ void Scene_Init(void) {
 
     gRestrictionFlags[72].flags3 = 0; // Allows farore's wind in GTG
     gRestrictionFlags[94].flags3 = 0; // Allows farore's wind in Ganon's Castle
+    // Allow Farore's Wind anywhere if the option is enabled
+    // Cannot be used in areas affected by a general item restriction
+    if (gSettingsContext.faroresWindAnywhere){
+        for (int i = 0; i < 99; i++){
+			gRestrictionFlags[i].flags3 &= ~0x30;
+		}
+    }
```
gamestabled: I think this code block was intentionally removed in a previous commit, but I dont want to say definitively until @HylianFreddy can double check

gamestabled (changes requested):

code/src/entrance.c
HylianFreddy: Yes that's correct. The FW anywhere setting is handled by a separate function, so changing the restriction flags isn't needed

PR #237, opened 2021-06-28
Ammo Drops and Heart Drops and Refills options

CharadeNut:
Adds Ammo Drops as an option, and merges the Bombchu Drops option to it:
• **On**: Behaves as normal;
• **On + Bombchu**: Adds Bombchu, works the same way the Bombchu Drops option did;
• **Off**: Ammo (seeds, arrows, bombs, bombchu, nuts and magic) drops are replaced with blue rupees. Ammo upgrades only replenish current ammo by 10.
The logic ensures that available shops will carry ammo refills and green potion so the player can never run out and lock progress. Big Fairies and Great Fairies also replenish magic as normal, but are not considered in the logic, because the player may need more than his normal magic gauge (e.g. when fighting Ganondorf).

Adds Heart Drops and Refills as an option:
**On**: Behaves as normal;
**No Drops**: Recovery Heart drops are replaced with green rupees, like when the player has full health in the base game. Fairy flex drops are still available as a failsafe.
**No Refills**: Health upgrades (Heart Containers, PoH, Double Defence) no longer refill the player's health upon being obtained. Fairies only heal the player by 3 hearts, including Big Fairies. The player has incentive to carry health-replenishing items like Milk and Red Potion, and refill health at Fairy Fountains.
**Off**: Enables both No Drops and No Refills.


cjohnson57:
Patches should not directly branch to C functions- it's unsafe. You need to add hooks in hooks.s, like how I showed you in discord. Although that was for that specific function, these other ones may be slightly different.

Also, the things in oot.ld should be in order of the addresses rather than tacked onto the end, and the order in patches.s and hooks.s should match those in oot.ld

Besides that, nothing else looks wrong at a glance- Sometime in the next few days I'll download your branch and try to ensure the logic is all working 100% correctly

code/oot.ld
```diff
@@ -954,6 +954,25 @@ SECTIONS
...
+		*(.patch_FairyPickupHealAmount)
+	}
+
+	.patch_FairyReviveHealAmount 0x34af48 : {
+		*(.patch_FairyReviveHealAmount)
+	}
+
+	.patch_FairyUseHealAmount 0x4968b4 : {
+		*(.patch_FairyUseHealAmount)
+	}
```
gamestabled: Can you move these around in this file to all be in order by address? Just for consistency with the rest of the file which is all in order.

We also usually Uppercase all of our hex numbers (I dont think this is 100% consistent in the repo, but its preferred)

code/src/item00.c
```diff
@@ -16,7 +16,30 @@
...
+        item00Type == 0x08 ||
+        item00Type == 0x09 ||
+        item00Type == 0x0A ||
+        item00Type == 0x0B ||
+        item00Type == 0x0C ||
+        item00Type == 0x0E ||
+        item00Type == 0x0F ||
+        item00Type == 0x10 ||
+        item00Type == 0x12 ||
+        item00Type == 0x19) {
```
gamestabled: I think it would be nice to use 
[this enum](https://github.com/zeldaret/oot/blob/3fbdccbdbac4113ae04d77216b2a28b405e07734/include/z64actor.h#L243) for this, it could probably just be put in item00.h
https://github.com/zeldaret/oot/blob/3fbdccbdbac4113ae04d77216b2a28b405e07734/include/z64actor.h#L243
It also might look better as a switch block like
```
switch(item00Type) {
    case ITEM00_BOMBS_A:
    case ITEM00_ARROWS_SINGLE:
    case ITEM00_ARROWS_SMALL:
        // ...
    case ITEM00_BOMBS_SPECIAL:
        item->actor.params = (item->actor.params & 0xFF00) | 0x01;

}
```

code/src/item_effect.c
```diff
@@ -13,7 +13,10 @@ void ItemEffect_None(SaveContext* saveCtx, s16 arg1, s16 arg2) {
 }
 
 void ItemEffect_FullHeal(SaveContext* saveCtx, s16 arg1, s16 arg2) {
-    saveCtx->healthAccumulator = 20 * 0x10;
+    //With the No Health Refills option on, store-bought health upgrades do not heal the player
+    if(!gSettingsContext.heartDropRefill == HEARTDROPREFILL_NOREFILL && !gSettingsContext.heartDropRefill == HEARTDROPREFILL_NODROPREFILL){
```
gamestabled: I think that there is errors in this condition due to missing parenthesis. You probably want it to be like
```
if((gSettingsContext.heartDropRefill != HEARTDROPREFILL_NOREFILL) && (gSettingsContext.heartDropRefill != HEARTDROPREFILL_NODROPREFILL))
```

code/src/item_effect.c
```diff
@@ -73,7 +76,10 @@ void ItemEffect_GiveSmallKey(SaveContext* saveCtx, s16 dungeonId, s16 arg2) {
 void ItemEffect_GiveDefense(SaveContext* saveCtx, s16 arg1, s16 arg2) {
     saveCtx->doubleDefense = 1;
     // saveCtx->defense_hearts = 20; //TODO? is this needed?
-    saveCtx->healthAccumulator = 20 * 0x10;
+    //With the No Health Refills option on, store-bought health upgrades do not heal the player
+    if(!gSettingsContext.heartDropRefill == HEARTDROPREFILL_NOREFILL && !gSettingsContext.heartDropRefill == HEARTDROPREFILL_NODROPREFILL){
```
gamestabled: same here

code/src/item_effect.h
```diff
@@ -18,6 +18,7 @@ void ItemEffect_GiveMagic(SaveContext* saveCtx, s16 arg1, s16 arg2);
 void ItemEffect_GiveDoubleMagic(SaveContext* saveCtx, s16 arg1, s16 arg2);
 void ItemEffect_GiveFairyOcarina(SaveContext* saveCtx, s16 arg1, s16 arg2);
 void ItemEffect_GiveSong(SaveContext* saveCtx, s16 questBit, s16 arg2);
+void ItemEffect_GiveUpgrade(SaveContext* saveCtx, s16 questBit, s16 arg2);
```
gamestabled: doesnt look the the name `questBit` is right. I would say either change it to just `arg1`, or rename `questBit` to `value` and `arg2` to `upgrade`.

code/src/patches.s
```diff
@@ -1271,6 +1271,31 @@ SariasSongHintsTwo_patch:
...
+FairyPickupHealAmount_patch:
+    bl FairyPickupHealAmount
+			   
+.section .patch_FairyReviveHealAmount
+.global FairyReviveHealAmount
+FairyReviveHealAmount_patch:
+    bl FairyReviveHealAmount
+			   
+.section .patch_FairyUseHealAmount
+.global FairyUseHealAmount
```
gamestabled: for consistency within this file this line should be
```
.global FairyUseHealAmount_patch
```
and similarly for the other four above this one

code/src/settings.c
```diff
@@ -56,6 +56,37 @@ s32 Settings_ApplyDamageMultiplier(GlobalContext* globalCtx, s32 changeHealth) {
 
     return modifiedChangeHealth;
 }
+//With the No Health Refill option on, full health refills from health upgrades and Bombchu Bowling are turned off, and fairies restore 3 hearts
+//Otherwise, they grant a full heal, and the default effect applies (full heal from bottle, 8 hearts on contact)
+u8 setFullHealthRestore(u8 setAmount) {
```
gamestabled: we usually name functions in the patch folder like `[SystemName]_[FunctionName]` (this isnt 100% consistent in the code folder but its nice) so for example this function would be called `Settings_SetFullHealthRestore`, and similarly for the ones below this one

gamestabled (changes requested):
Yooo Charade great job with this, I know you've been learning super fast and working hard on this feature, and I think a decent amount of players will enjoy it!

I left a few comments, most of which are nitpicky stylistic stuff. We should be able to get this merged in to be in the nightlies after they get addressed. lmk if you have any questions

gamestabled:
I just saw ShadowShine's comment too, sounds like he saw some potential (and valid) concerns with the patches. Those should be addressed and tested as well

code/src/item00.c
```diff
@@ -14,29 +14,60 @@
...
+    /* 0x16 */ ITEM00_SHIELD_HYLIAN,
+    /* 0x17 */ ITEM00_TUNIC_ZORA,
+    /* 0x18 */ ITEM00_TUNIC_GORON,
+    /* 0x19 */ ITEM00_BOMBS_SPECIAL
+} Item00Type;
+
 void EnItem00_rInit(Actor* thisx, GlobalContext* globalCtx) {
     EnItem00* item = THIS;
-    s16 item00Type = item->actor.params & 0x00FF;
+    s16 Item00Type = item->actor.params & 0x00FF;
```
gamestabled: i think it could be confusing to have this variable use the same name as the enum. i would go with either `item00Type` or `dropType` for the variable name

gamestabled (changes requested):
Looks good to me besides the one new comment I left, which is minor and a quick change. Nice job fixing this up

source/logic.cpp
```diff
@@ -175,10 +175,17 @@ namespace Logic {
...
 
   bool BuyBombchus5     = false;
   bool BuyBombchus10    = false;
   bool BuyBombchus20    = false;
+  bool BuySeed          = false;
+  bool BuyArrow         = false;
+  bool BuyBomb          = false;
+  bool BuyGPotion       = false;
+  bool BuyBPotion       = false;
+  bool MagicRefill      = false;
```
cjohnson57, line 188: I just realized- You forgot to add these variables to LogicReset(). All you have to do is put them there in the same order they're in here and set them to false.

gamestabled (approved):

cjohnson57 (approved):
There are some slight issues with this due to old code not taking the setting into account- After this PR is merged I'll momentarily make another one fixing these issues. Besides that, all seems good!

PR #238, opened 2021-06-28
Settings: Shuffle Medigoron and Carpet Salesman

HylianFreddy:
- This setting shuffles the Giant's Knife from Medigoron and the 10 Bombchus pack from the Haunted Wasteland Carpet Salesman into the randomization pool, like it does in OoTR. When it is selected, the merchants will sell their random item only once.
- For the players that don't like spending 200 rupees on junk, an extra option will make the merchants tell you what they're actually selling, using the same text that appears in the gossip stone hints.
- Blue Potions are available to buy at Granny's shop after obtaining the Claim Check (or any trade item after the Odd Mushroom, for when trade item randomization will eventually be implemented).


gamestabled:
What happens if you already have the biggoron sword and then get giant's knife? if it downgrades, we gotta prevent that. we could potentially make them progressive when medigoron is shuffled, i wonder if people would like or dislike that.

besides this question, looks good to me, and the hints are sick

HylianFreddy:
Getting Giant's Knife doesn't downgrade the Biggoron's Sword.
I personally don't think we should make the swords progressive, because the Giant's Knife is little more than junk, so it acts more like an Ice Trap (imagine seeing it from a distance and hoping it's the Biggoron's Sword).
Also I haven't yet found how to make the Biggoron's Sword model actually appear, I'll have to look into that too.

PR #246, opened 2021-06-30
Only show a dungeon's reward when it's compass is acquired

Kewlan:
~~**Based on #245, merge/close that first**~~

image
In OoTR this is an option, but I think it's good to keep the options low when possible and, again, I'm guessing that most are fine with this always enabled.

I also added that extra message, but maybe that's too much?


Kewlan:
I forgot that the locations are always written at the altar in Temple of Time, so this might be pointless. Or it can be seen as the compass becomes more useful because it let's you see the hint anytime.

gamestabled:
I personally think that this needs to be an option. When I've played OoTR in the past, I always made it so I didnt need the compasses

Kewlan:
In that case I'll opt to just close this since the player can always see the locations at the altar. It'd probably just be an annoyance anyway.

I'll close this pull request tomorrow if there are no objections by then.

PR #250, opened 2021-07-02
Fix issues with in-game spoiler log data and item collection check data

Giometric:
This PR fixes two issues with the in-game spoiler log and item list:

- Kakariko Bazaar item collection check data now has the correct values; the in-game log and item list will now correctly highlight when those shop items have been purchased.
- When spoiler data is set to "hidden", playthrough data filling is now skipped entirely. This avoids an issue that can occur when items that were skipped from adding to the items list were still considered part of one of the playthrough spheres.


Kewlan (approved):
Tested and works! 👍

PR #253, opened 2021-07-03
Fix Progressive Bombchu shop text and exclude from item tracker

Kewlan:
As mentioned on Discord, before this it said they were a one time purchase but were re-purchasable anyway. So this fixes the text and tracker until-and-if it's decided that they should only be purchasable once.
image

Also excludes the minigame bombchu drops and excluded locations from the item tracker when spoilers are off.
image


Kewlan:
New commit: Should now exclude all possible uncheckable locations regardless of settings, excluding Ganon. The aim is to make it possible to make the entire tracker page green for those who like to 100%, and clears up space of pointless checks.

PR #255, opened 2021-07-05
Random Mweeps & Starting Health

Roberto-Nessy:
Added settings Fast (current behaviour), Vanilla, and Random (anywhere up to 128 mweeps with an average around 10~20) for King Zora Speed. (1 in 65536 chance of getting all 128 mweeps)

Also added health to the starting items menu with options from 3 to 20 hearts. This will remove heart containers from the pool followed by pieces of heart 4 at a time.
The code is also ready to allow 1 and 2 hearts (which replace 4 pieces of heart with 2 or 3 containers), however the logic for this has not been implemented so those options have not been added yet.


Roberto-Nessy:
Saw people in #general talking about how double defense diminished the challenge of double/quad damage so I added a setting to remove it. Also locks StartingDoubleDefense off since it wouldn't make sense to remove it and also start with it.

PR #257, opened 2021-07-07
Settings: Add closed deku setting

cjohnson57:
Adds a setting for closed deku, which makes it so that the Kokiri Sword and Deku Shield are required to get past Mido and get to Deku Tree, but you can still get past the Kokiri boy and leave the forest for free.

I tested to ensure the correct entrances were or weren't guarded depending on the setting.

Logic was actually already written in such a way that it didn't need to be modified for this setting:
* For Deku Tree access, it checks for open forest, and since closed deku isn't open forest logic correctly concludes it can't be reached for free.
* For being able to leave Kokiri Forest, it checks that the setting is not closed forest, and since closed deku isn't closed forest it correctly concludes you can leave forest for free.


gamestabled:
In the base game, Mido stops blocking the deku tree once you have visited zelda in the courtyard. Does that still happen and need to be addressed for this setting?

cjohnson57:
Thanks for the heads up, didn't know about that, it is indeed an issue, so I'll have to figure out something to do about that.

PR #279, opened 2021-07-23
Add "Intended" Door of Time option

Kewlan:
When set to "Intended", the Door of Time will only open when the Song of Time is played on the Ocarina of Time while having all three spiritual stones.

~If it matters, the "correct song played" tune still plays when playing the Song of Time without the necessary requirements.~


cjohnson57 (approved):
Looks good to me

Kewlan:
New commits from Freddy:
- Playing SoT to open the DoT with the "intended" setting will play the error sfx when the door can't be opened
- The spiritual stones on the ToT pedestal will only appear if the DoT is open, even for adult Link

gamestabled (commented):
Everything here looks good to me, but I have one more request: Can we edit the child altar text depending on this setting? Right now it always says `"Ye who may become a Hero...&Stand with the Ocarina and&play the Song of Time."`

source/hints.cpp
```diff
@@ -501,6 +501,26 @@ static Text BuildDungeonRewardText(ItemID itemID, const ItemKey itemKey) {
   return Text()+ITEM_OBTAINED(itemID)+"#"+GetHintRegion(Location(location)->GetParentRegionKey())->GetHint().GetText()+"#...^";
 }
 
+static Text BuildDoorOfTimeText() {
+  std::string itemObtained;
+  Text doorOfTimeText;
+
+  if (OpenDoorOfTime.Is(OPENDOOROFTIME_OPEN)) {
+    itemObtained = ITEM_OBTAINED(ITEM_SWORD_MASTER);
```
Kewlan, line 509: I don't know if there's a way to remove the spacing for the image; When I kept `itemObtained` empty it used the previous image, and when I set it to `ITEM_NONE` it kept the spacing with no image. So I set it to the Master Sword as it is the path to the future.

gamestabled: what if in that case you just return `Text() + doorOfTimeText` without the itemObtained?

Kewlan: Then it still reuses the previous image like so:
image
image

gamestabled: gotcha. then master sword image sounds good for now. If there is ever some time of "shuffle master sword" in the future, we can come back to it

PR #280, opened 2021-07-24
Add "Progressive Goron Sword" option

Kewlan:
When enabled, a Giant's Knife will always be found before Biggoron's Sword, and Medigoron won't start selling knives until the GK has been found in the wild and broken.
Also allow Giant's Knives to be bought after acquiring the shuffled check.

One thing to note is that with how I made the patch that chooses the custom text, Medigoron's `I just completed a small weapon. How about it...` message is now skipped since that leads to 304F. I feel that the amount of code required to make it work is not worth it for just this one message. Perhaps in the future my `GetCustomText` patch can be removed if the ability to insert variables into text messages gets added. Then the text at 304F can be replaced instead.

This should probably be tested a bit before merging since there are several combinations of options that affect stuff here, with Shuffle Merchants being Off, On (No Hints), or On (Hints), and Progressive Goron Sword being Off or On.


Kewlan:
Latest force push was just adding a missing `\n` in the settings description.

PR #284, opened 2021-07-26
Settings: Quick Text and Skip Song Replays

HylianFreddy:
New Quick Text setting:
- "Skippable" makes every text box instantly complete when you press the B button.
- "Instant" makes every text box print out in full, without scrolling.
- "Turbo" lets you advance text by holding B instead of mashing it.

New Skip Song Replays setting:
- After you play the last note of a song, it won't automatically replay and you will immediately get control of Link (or trigger whatever cutscene comes afterwards in that situation).

There are still a couple of pretty minor problems that could be resolved to improve the experience, but I've opened the PR anyway so I can get some feedback.

Notes on how the code works:
- Skippable text: The "Unskippable" control code is skipped over.
- Instant text:   For every text sequence, the "Instant On" control code is run once when the first character is parsed, and also each time the "Wait For Input" one appears. The "Instant Off" control code is skipped over.
- Turbo text:     The check to see if B is pressed for advancing the text is replaced with the randomizer check on the same button, so it returns true even if the button has never been released.
- Skip Replays:   Simply changing the value of msgMode is enough to skip the replay, but the Ocarina Effect actors also need to be killed as soon as they spawn to prevent glitches. This can't be done with the Sun's Song light and the Song of Storms rain, as they're responsible for most of those songs effects. I fixed the light by calling one of its own actor functions, and I just left the rain actor unchanged.

Known issues:
- Turbo text doesn't close text boxes, so the text can still stop advancing if it reaches a GOTO control code.
- ~Turbo text doesn't work if you hold other buttons alongside B.~
- ~Song of Storms grottos open up after a certain timer in the rain actor counts down to 0x134. I haven't found this timer yet, so you have to wait for about 5 seconds of raining for the grotto to open.~
Thanks to Gamestabled, these 2 issues were fixed


code/src/settings.c
```diff
@@ -94,6 +96,24 @@ void FairyPickupHealAmount(void) {
     }
 }
 
+u32 Settings_GetQuickTextOption() {
+    return gSettingsContext.quickText;
+}
+
+u32 Settings_IsTurboText() {
+    return (gSettingsContext.quickText >= QUICKTEXT_TURBO && buttonCheck(BUTTON_B));
```
gamestabled: I think instead of using `buttonCheck` you might just want to check `(rInputCtx.cur & BUTTON_B)`. Haven't tested, but i think that will fix the issue of being able to hold other buttons, and i think `buttonCheck` is meant to stay internal to `input.c` anyway

HylianFreddy: Done. As you predicted, this fixed that issue. Thank you

PR #289, opened 2021-07-29
Preperations for Glitched logic options

Kewlan:
Before I do more work on this I want to hear if this looks good.
Things might still be broken, but it builds and can generate the mod.

- Turn the ConfitionFn to a vector where the condition iterates through the easiest difficulty to the current set option.
- Add basic easy glitched logic bools.
- Add a few examples in location_access.cpp.

Some examples:
image
image
image
image

When the Logic setting is set to Glitched Easy or Glitched Hard it first checks if the easier conditions are met, other wise it checks the next difficulty until it reaches the set difficulty option.

With this way, we won't have to repeat a lot of code!

### DIFF
--- a/code/src/settings.h
+++ b/code/src/settings.h
@@ -15,6 +15,7 @@ typedef enum {
 
 typedef enum {
   LOGIC_GLITCHLESS,
+  LOGIC_GLITCHED,
   LOGIC_NONE,
   LOGIC_VANILLA,
 } LogicSetting;
--- a/source/entrance.hpp
+++ b/source/entrance.hpp
@@ -29,8 +29,28 @@ enum class EntranceType {
 class Entrance {
 public:
 
-    Entrance(AreaKey connectedRegion_, ConditionFn conditionsMet_)
-        : connectedRegion(connectedRegion_), conditionsMet(conditionsMet_) {}
+    Entrance(AreaKey connectedRegion_, std::vector<ConditionFn> conditions_met_)
+        : connectedRegion(connectedRegion_) {
+        conditions_met.resize(2);
+        for (size_t i = 0; i < conditions_met_.size(); i++) {
+            conditions_met[i] = conditions_met_[i];
+        }
+    }
+
+    bool GetConditionsMet() const {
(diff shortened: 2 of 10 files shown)


cjohnson57:
The overall idea for this is good. It's probably a lot better to do it this way than to have separate location access files for glitched logic. And the way conditions are processed seems good. Since for the purposes of ConditionsMet() Vanilla logic is equivalent to none, I think these changes for how conditions are processed don't even need to be updated for that PR.

That being said, is having GlitchedEasy and GlitchedHard the approach we want to take? The currently underway glitch logic rewrite for OoTR is instead trickified- Where base glitched logic would be like what your GlitchedEasy is, and harder glitches are handled the way tricks are. So in your first example, you might have like `(IsAdult && CanHover) || (LogicReallyHardGlitch && CanDoReallyHardGlitch)`. Of course, this approach would be more difficult than your idea for just having Easy and Hard logic, so it's also a question of how much free work the person writing glitched logic wants to do, so although I'll mention this idea ultimately it'll be up to that person (and based on this PR it seems that person is you :p)

Kewlan:
> It's probably a lot better to do it this way than to have separate location access files for glitched logic.

Yes, and for any glitchless checks that always return true we don't need to do anything, as Glitched will first check the Glitchless check.

> That being said, is having GlitchedEasy and GlitchedHard the approach we want to take? The currently underway glitch logic rewrite for OoTR is instead trickified- Where base glitched logic would be like what your GlitchedEasy is, and harder glitches are handled the way tricks are. So in your first example, you might have like `(IsAdult && CanHover) || (LogicReallyHardGlitch && CanDoReallyHardGlitch)`. 

This sounds better. We can have a submenu appear below Glitched when it's selected with the tricks inside.

Kewlan:
Commit updated:
image
image
image

Kewlan:
Update: Glitches can now be set to only use certain difficulties. Some examples would be Beginner plus Intermediate, only Beginner, Intermediate and Advanced. Every glitch can still be disabled.

PR #306, opened 2021-08-06
Redefining setting descriptions

Kewlan:
- Define descriptions directly in the header file, don't use extern.
- Remove the cpp file.
- Rename alias.

May or may not fix the GlitchTrick description issue. Even if it doesn't, we may want this as it is one less file to deal with.


gamestabled:
I dont think we want to merge this, actually. We probably don't want to define the descriptions directly in the header, because that could cause problems in the future if we need another file to include the header. I also wouldn't expect this to solve the issue with GlitchTrick (and if it does, i think its just out of coincidence at the moment; it could break again because it is still undefined behavior).

Kewlan:
> ...because that could cause problems in the future if we need another file to include the header.

That's what `#pragma once` is for, right? And these descriptions are just constants that doesn't rely on anything else, so it should be fine.

> I also wouldn't expect this to solve the issue with GlitchTrick (and if it does, i think its just out of coincidence at the moment; it could break again because it is still undefined behavior).

If the issue is that the vectors sometimes aren't defined yet, then wouldn't this solve the problem since they get defined as soon as they're declared?

Kewlan:
Nevermind, I don't know C++.

PR #307, opened 2021-08-06
Add and run style format script

cjohnson57:
Adds binaries, scripts, and settings for Artistic Style in order to clean up the repo and keep our code style consistent. Uses the following settings (explained in more detail [here](http://astyle.sourceforge.net/astyle.html))

`--style=java`
Uses our java style brackets, where opening brackets are on the same line and closing brackets are on their own line.
`--indent=spaces=4`
Uses 4 space indentation for all files.
`--indent-switches`
`--indent-cases`
`--indent-namespaces`
`--indent-preproc-block`
`--indent-preproc-cond`
Just makes sure the aforementioned things add spacing.
`--pad-oper`
`--pad-comma`
`--pad-header`
Adds spacing between operators in arithmetic statements, after commas, and after headers (ex. if() -> if ())
`--align-pointer=type`
`--align-reference=type`
Ensures that \*'s and \&'s are aligned to the type (ex. int* x)
`--add-braces`
Adds braces to single-line if statements, loops etc.
`--convert-tabs`
Ensures there are no tabs used for spacing.
`--close-templates`
Not really an issue but removes extra spacing between <>'s in templates.

I ran the formatter and went through each file, and as far as I can tell nothing was messed up. We have a lot of files with tables spaced a very specific way for readability and I was worried it would screw these up, but it seems to not be the case, except in settings_descriptions.cpp, hint_list.cpp, and location_access.cpp which is why those files are excluded in the settings for cpp files. I also excluded all files which, as far as I can tell, were not created by us, which includes tinyxml and several .c files (for the full list see [the c style options](format/style-options-c.ini))

I also added a section to the readme with instructions for running the formatter.


lioncash:
Is there a particular reason clang-format wasn't used instead of astyle?

It comes packaged alongside most IDEs (e.g. Visual Studio has native support for it, as does VSCode and includes hooks by default for them, among other things), which makes for easier integration for the average dev. This also means binaries for the formatting program don't need to be included in the repo, allowing the user to use their system-provided version of the formatter (or IDE provided version). It also means repo maintainers don't need to keep the executables updated either.

cjohnson57:
I didn't want to really require any extra setup from contributors. Even if VSCode comes with it clang-format still seems to require some setup (at least, even if it came installed I don't seem to have it ready to go on my PC). With this they can just run the scripts and not have to worry about it.

>  This also means binaries for the formatting program don't need to be included in the repo, allowing the user to use their system-provided version of the formatter 

Is this a good thing? Couldn't that potentially cause discrepancies?

>  It also means repo maintainers don't need to keep the executables updated either.

I'm not really worried about this as I don't think we'll need to unless a new version comes out with an option we want that didn't exist previously

cjohnson57:
Update: It was decided the effect on the formatting of location_access and hint_list broke formatting too badly so they were excluded.

cjohnson57:
Update: It was decided to standardize spacing to 4 spaces per indent across all files. This also made it unnecessary to have separate options files for .hpp and .cpp files

gamestabled:
I think lioncash could be right here that we would prefer to use clang-format.
> 
> I didn't want to really require any extra setup from contributors. Even if VSCode comes with it clang-format still seems to require some setup (at least, even if it came installed I don't seem to have it ready to go on my PC). With this they can just run the scripts and not have to worry about it.
> 
This is fair that it requires more setup, but contributing already (in theory) requires the devkitARM toolchain, which is a much more serious and less common prerequisite.
> > This also means binaries for the formatting program don't need to be included in the repo, allowing the user to use their system-provided version of the formatter
> 
> Is this a good thing? Couldn't that potentially cause discrepancies?

Avoiding binaries in the repo (especially ones from external materials, even if we have it licensed properly) is probably a better trade-off than minimal clang discrepancies. And really I don't think that it would be an issue, since clang-format is so commonly used.
> > It also means repo maintainers don't need to keep the executables updated either.
> 
> I'm not really worried about this as I don't think we'll need to unless a new version comes out with an option we want that didn't exist previously

Well, staying up to date on executables isn't just about feature updates, but also bug fixes (esp. security vulnerabilities).

Also, here's another argument in favor of clang-format: The style that I have used in my code has been based off the style guide for [the oot decomp project](https://github.com/zeldaret/oot), so we could make that consistency "official" by just stealing their configurations. The decomp repo is quickly becoming (or, already is) the de facto oot reference source, and I bet when OoTR eventually gets ported to decomp, they will keep the same style. So basically, this would keep us consistent with closely related projects.

cjohnson57:
Those are fair points, however (and I hope this doesn't come off as belligerent) I don't really feel like figuring out clang-format and redoing all of this in it so if that's what we want to do then someone else will have to do it

cjohnson57:
Also does clang-format work/exist on Windows? I can't find any solid word that it does

lioncash:
Yup, it does.

All that's necessary to support the same subset of oot decomp is to drop their clang-format file into the project root and run it using that config file (where we can amend the format file if necessary).

PR #315, opened 2021-08-16
Get difficulty options from vector function to avoid accidental misnaming

Kewlan:
With this, when more glitches are added we only have to check that the `glitchDifficulties` array is the same as the vector in `setting_descriptions.cpp`.

### DIFF
--- a/source/settings.cpp
+++ b/source/settings.cpp
@@ -594,14 +594,29 @@ namespace Settings {
     &LogicSpiritTrialHookshot,
   };
 
-  Option GlitchISG = Option::U8("Infinite Sword Glitch", {"Disabled", "Novice"}, {GlitchISGDescDisabled, GlitchISGDescNovice});
-  Option GlitchHover = Option::U8("Bomb Hover", {"Disabled", "Novice", "Intermediate", "Advanced"}, {GlitchHoverDescDisabled, GlitchHoverDescNovice, GlitchHoverDescIntermediate, GlitchHoverDescAdvanced});
-  Option GlitchMegaflip = Option::U8("Megaflip", {"Disabled", "Novice", "Intermediate"}, {GlitchMegaflipDescDisabled, GlitchMegaflipDescNovice, GlitchMegaflipDescIntermediate});
-  Option GlitchHookshotClip = Option::U8("Hookshot Clip", {"Disabled", "Novice"}, {GlitchHookshotClipDescDisabled, GlitchHookshotClipDescNovice});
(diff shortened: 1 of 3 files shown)


source/settings.cpp
```diff
@@ -594,14 +594,29 @@ namespace Settings {
...
-  Option GlitchHookshotClip = Option::U8("Hookshot Clip", {"Disabled", "Novice"}, {GlitchHookshotClipDescDisabled, GlitchHookshotClipDescNovice});
-  Option GlitchHookshotJump_Bonk = Option::U8("Hookshot Jump (Bonk)", {"Disabled", "Intermediate"}, {GlitchHookshotJump_BonkDescDisabled, GlitchHookshotJump_BonkDescIntermediate});
-  Option GlitchHookshotJump_Boots = Option::U8("Hookshot Jump (Boots)", {"Disabled", "Novice", "Intermediate"}, {GlitchHookshotJump_BootsDescDisabled, GlitchHookshotJump_BootsDescNovice, GlitchHookshotJump_BootsDescIntermediate});
-  Option GlitchLedgeClip = Option::U8("Ledge Clip", {"Disabled", "Novice", "Intermediate"}, {GlitchLedgeClipDescDisabled, GlitchLedgeClipDescNovice, GlitchLedgeClipDescIntermediate});
-  Option GlitchTripleSlashClip = Option::U8("Triple Slash Clip", {"Disabled", "Novice"}, {GlitchTripleSlashClipDescDisabled, GlitchTripleSlashClipDescNovice});
+  //Function to avoid accidentally naming the options wrong, as logic.cpp requires these exact names
+  const std::vector<std::string> GlitchDifficultyOptions(u8 enabledDifficulties) {
```
lioncash:
```suggestion
  std::vector<std::string> GlitchDifficultyOptions(u8 enabledDifficulties) {
```

source/settings.cpp
```diff
@@ -594,14 +594,29 @@ namespace Settings {
...
+    const std::vector<std::string> glitchDifficulties{"Novice", "Intermediate", "Advanced", "Expert", "Hero"};
```
lioncash:
```suggestion
    static constexpr std::array glitchDifficulties{"Novice", "Intermediate", "Advanced", "Expert", "Hero"};
```

Would allow this to work without constructing the string on every function invocation

Kewlan:
New commit: Rewrite conditions met checks to be more readable. 
The old way was made for when it was `Glitched Easy` and `Glitched Hard`, and if someone wanted to add an even harder difficulty. This new style makes more sense for what we have now, and works better if there will ever be another logic option that, for example, wants to only check a newly added `conditions_met[2]`, or `[0]` first and then `[2]`, or whatever.

PR #331, opened 2021-08-30
Add "Keep Farore's Wind Warp Point" option

Kewlan:
Also enable "Picked up Deku Seeds" bit

### DIFF
--- a/code/oot.ld
+++ b/code/oot.ld
@@ -692,6 +692,14 @@ SECTIONS
 		*(.patch_FairyReviveHealAmount)
 	}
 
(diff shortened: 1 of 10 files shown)


gamestabled:
Is this non-optional as you've made it here? And if it's turned on, how do remove and set a new FW point?

Kewlan:
It is indeed non-optional. To remove the one that's currently set you just select Dispel in the menu. Opening the menu takes just a portion of a second, and dispelling is instant.
image

gamestabled:
In my imo, I think this feature needs to be optional, probably under the time savers menu. I feel weird about pushing a change like this without it being an option

PR #334, opened 2021-08-31
Add in-game playtime timer, in-game menu improvements

Giometric:
This PR adds a total playtime timer to the in-game menu, shown in the first tab which shows the seed. The timer ticks up whether the menu is open or closed, but does not tick up if the system is asleep. The playtime is saved to the user's file (using a hopefully-unused spot in the save data).
New drawing methods were added to the draw.c file, using an extra buffer which acts as a 'backbuffer' to draw to. The backbuffer is copied to both of the system framebuffers at the end of rendering without clearing the previous framebuffer contents, to remove flashing when the user switches screens or scrolls the item lists. A future improvement to this could include finding out which buffer is in front, so the backbuffer can be copied only once to that one, saving a significant amount of time.
An attempt at enabling -O1 for the C-side code was made to speed up the code in the in-game menu, but the optimizations (seems to be inlining specifically) appear to cause issues with some of the patch code, so it was disabled again.
Finally, hooks were added into the game's handlers for system sleep and awake events, and the menu responds to these by breaking out of its draw loop so that the game can fully respond, and the system can go to sleep as normal. The warning message about putting the system to sleep while in the in-game menu was removed, as it no longer applies.


code/oot.ld
```diff
@@ -8,6 +8,14 @@ SECTIONS
...
 	}
 
+	.patch_SleepQueryCallback 0x3FD6C4 : {
+		*(.patch_SleepQueryCallback)
+	}
+
+	.patch_AwakeCallback 0x3FD43C : {
+		*(.patch_AwakeCallback)
+	}
+
```
gamestabled: Could you move these to be sorted by address? Just for consistency with the rest of the file.

gamestabled (changes requested):
Very cool PR. The sleep mode thing has been an issue since before the randomizer was a randomizer, and the other graphics fixes/improvements are also good stuff. I just left one comment about a consistency thing, and the rest looks good to me.

code/oot.ld
Giometric: I somehow managed to completely not notice the patches were ordered by address in that file! Fixed in the last push.

gamestabled (approved):

PR #339, opened 2021-09-02
Add "Shuffle SFX" option

Kewlan:
Comes with 4 options:
- Off
- All
- Scene Specific
- Chaos

And another that can be set separately:
- Categorical Shuffle


Kewlan:
As can be seen in `sfx.h`, I made several arrays of sound effects even though they share a lot of the same IDs. I thought it'd be better to save some computing time at the cost of some space instead of getting a list of the right sound effects every time one is going to be played since that sometimes happens several times each frame.

PR #349, opened 2021-09-14
Unshuffle common sound effects that were broadly categorised

Kewlan:
These sound effects would often become something that doesn't fit at all and would sound terrible.


Kewlan:
I'll make another PR where I recategorise more sound effects.

PR #351, opened 2021-09-17
Fix Lost Woods music playing after exiting Goron City in Overworld ER

### LINKED ISSUE #277: Lost woods music issues with overworld entrances randomizer
Using the v2.0-4e7748 nightly version:
- On the entrances that normally lead to the LW bridge (KF->bridge or HF->bridge), the music is set to nature noises regardless of what the music of the destination map is.
- On the GC->LW entrance, the LW music plays on the destination map on top of the normal music until the normal music changes eg. because of the day/night cycle

gymnast86:
The Lost Woods music that plays near the GC Woods Warp -> Lost Woods exit would continue to play over the next areas music when Overworld Entrance Randomizer was enabled. This is fixed by setting the fadeout animation of the exit to be a Lost Woods fadeout.

### DIFF
--- a/code/src/entrance.c
+++ b/code/src/entrance.c
@@ -168,6 +168,15 @@ void Entrance_Init(void) {
             }
         }
     }
+
+    //Set the exit transition of GC Woods Warp -> Lost Woods to a lost woods transition.
+    //This works as an easy fix for the Overworld ER bug that continues to play the lost
+    //woods music into the next area, even if isn't the lost woods. A "proper" fix would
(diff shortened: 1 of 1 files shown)


gamestabled:
Could we make this only happen if music shuffle is on? Not a big deal but seems nice to reduce the scope of the change

UraYukimitsu:
The issue happens even without music shuffle on overworld ER

gamestabled:
Yeah, I meant to say overworld ER, not music shuffle. Thanks for the clarification lol

PR #361, opened 2021-09-20
Fix the final two build errors

Kewlan:
Fixes these two build errors that has existed for how ever long:
image

I don't feel fully confident about the one in `item_effect.c`, but if I've understood correctly the parameter is still declared correctly. The functionality in-game also still works.


cjohnson57:
The one in item_effect should be good yeah

PR #365, opened 2021-09-22
Place the excluded locations options into grouped-by-area submenus

Kewlan:
Also stores a `settingBound` in each menu to fix the current behaviour where scrolling in one menu would scroll in all others.


Kewlan:
At first I was planning to define `excludeLocationsMenus` with a lambda function, but because of potential issues like undefined behavior I'm thinking of waiting with that until and if this is merged.

Kewlan:
The latest commit fixes an issue where you could change a hidden option by selecting it with the menu cursor, and then making the app hide it. Changing that hidden option softlocks the app as shown below:

https://user-images.githubusercontent.com/5352197/134395414-d9b7dcc3-5c07-4953-a9bc-faa12fc01c47.mp4

Shuffle Kokiri Sword gets turned off while the Kokiri Sword Chest location is hovered in that menu.

It was possible to have a hidden option selected by resetting to default or loading a preset that hides the hovered option. It's likely the same issue would happen with a locked option, but this should fix some cases of that too.

PR #367, opened 2021-09-24
Add scene and entrance discovery tracking

Giometric:
This PR adds basic scene and entrance 'discovery' tracking to the in-game menu. Whenever a map is about to be loaded, the hook checks which scene number and entrance ID have just been entered, and sets the appropriate bits in extData to remember which scenes and entrances the player has seen in that playthrough.
Currently this is used only to help finish up the dungeon info page's tracking of map modes. A dungeon's mode (Master Quest or Vanilla) is revealed either by collecting the map for the dungeon, if the option is enabled in the app, or by visiting that dungeon at least once. The modes for Ganon's Castle and Gerudo Training Grounds, which do not have maps, can only be revealed by visiting them.
The entrance tracking is not currently used for anything, but may be used in the future to supplement entrance randomizer playthroughs by allowing for built-in tracking of which randomized entrances the player has seen.


code/src/savefile.c
```diff
@@ -236,6 +236,47 @@ u8 SaveFile_GetDungeonCount(void) {
...
+void SaveFile_SetEntranceDiscovered(u16 entranceIndex, u8 isAdult, u8 isNight) {
+    // Most entrances have 4 'slots':
+    // base   - Child Day
+    // base+1 - Child Night
+    // base+2 - Adult Day
+    // base+3 - Adult Night
+    // Based on the values of isAdult and isNight, entranceIndex will be moved to the base index,
+    // and then all four slots will be marked as discovered
+    if (isNight) { entranceIndex -= 1; }
+    if (isAdult) { entranceIndex -= 2; }
```
HylianFreddy: This is not necessary, `entranceIndex` is already the base value. The game never saves the "entrance with offset" anywhere afaik, so when you see a struct in the Save Context containing an entrance index it is always the base one.
For example if Adult Link enters Ganon's Castle from the Market this variable will be 0x138. The actual entrance the game used to load the scene is 0x13A, but that's never saved, it is only used for the calculations during the load.

code/src/savefile.c
```diff
@@ -236,6 +236,47 @@ u8 SaveFile_GetDungeonCount(void) {
...
+
+    u16 numBits = sizeof(u32) * 8;
+    for (u32 i = 0; i < 4; i++) {
+        u32 entranceIdx = entranceIndex / numBits;
+        u32 entranceBit = 1 << (entranceIndex - (entranceIdx * numBits));
+        gExtSaveData.entrancesDiscovered[entranceIdx] |= entranceBit;
+        entranceIndex++;
+    }
```
HylianFreddy: This is just a thought, but it feels kinda weird to save each individual entrance like this: you count all 4 entrances as visited when only one of them is triggered, so why not ignore the offset ones and just count the base as "1 entrance"? My suggestion would be to ditch the entrance index and use the spawn index returned by the entrance table, which maybe could also work better for Entrance Randomizer.
For example, if you exit Lost Woods towards Hyrule Field the base `entranceIndex` will be 0x185, even with ER active. You could then access the entrance table and save the spawn index as "visited": without ER that is spawn 03, which places Link near the LW exit; if ER is active and you end up, say, at "Gerudo Fortress from wasteland" then the spawn will be 0F, but you can save it the same way (because ER overwrites the entrance table, so entrance 0x185 will have scene 5D and spawn 0F in this example)

Giometric: After poking around some more, since there are much fewer ER overrides than there are entrances, and since the entrances should be unlocked in 'pairs' anyway, what I may do is send a little more data to the patch so that I can figure out which pair, if any, a visited entrance belongs to, and then track completions on the pairs just by index into an array of those. Or, track the entrances like I'm doing (just the base ones, since as you mentioned the game resolves to the base index anyway by the time an area is being loaded), then list each pair, and check if either of the pair's entrances have been visited. This way wastes a small amount of memory, but it's still only around 200 bytes total to keep a list of all 1555 entrances - the string data for all the location names will easily eclipse that on its own.
It is mild spoilers to be able to find out how many different entrance pairs have been randomized with the list, but I do plan to allow the user to turn off ER tracking if they want to.

PR #377, opened 2021-10-06
Fix some minor ER issues

gymnast86:
- Properly grouped the regular interiors and special interiors together when Shuffle Interiors is set to All
- Fixed a rare randomization case that would result in infinite recursion
- Changed the exit of DMT_COW_GROTTO to be correct


lioncash (approved):
swag

PR #378, opened 2021-10-07
Add entrance pair tracking with names

Giometric:
This PR finishes up the basic entrance tracking functionality. The "Randomized Entrances" tab in the in-game menu will now display a list of hidden and discovered entrance pairs. These are revealed and colored green once the player has traveled through either of the entrances in the pair.
image
Data for the pairs, including all the entrance names, is created by the app and sent over in the patch. The names themselves still require touching up and the implementation may need to be refined later to allow localization of the entrance names.


code/src/gfx.c
```diff
@@ -377,36 +377,46 @@ static void Gfx_DrawSpoilerItemGroups(void) {
...
+                }
+            }
+
             u32 color = isDiscovered ? COLOR_GREEN : COLOR_WHITE;
-            Draw_DrawFormattedString_Small(10, locPosY, color,
-                "0x%04X -> 0x%04X, BW: 0x%04X", entrOverride.index, entrOverride.override, entrOverride.blueWarp);
+            const char* unknown = "???";
+            const char* startName = pair.StartStrOffset != ENTRANCE_INVALID_STRING_OFFSET ? &gEntranceTrackingData.StringData[pair.StartStrOffset] : "START STRING NOT FOUND";
+            const char* returnName = pair.ReturnStrOffset != ENTRANCE_INVALID_STRING_OFFSET ? &gEntranceTrackingData.StringData[pair.ReturnStrOffset] : "RETURN STRING NOT FOUND";
+            Draw_DrawFormattedString_Small(10, locPosY, color, "%s ->", isDiscovered ? startName : unknown);
```
Kewlan:
```suggestion
            Draw_DrawFormattedString_Small(10, locPosY, color, "%s <->", isDiscovered ? startName : unknown);
```
Since entrances are always coupled and the list doesn't have an entry for the entrance on the 2nd row.

code/src/gfx.c
```diff
@@ -377,36 +377,46 @@ static void Gfx_DrawSpoilerItemGroups(void) {
 
 static void Gfx_DrawERTracker(void) {
```
Kewlan, line 394: This menu should be hidden when entrance randomization is disabled, which is done for other menus around line 613.

code/src/gfx.c
```diff
@@ -377,36 +377,46 @@ static void Gfx_DrawSpoilerItemGroups(void) {
...
+        u16 itemCount = gEntranceTrackingData.EntrancePairsCount;
         u16 firstItem = entranceScroll + 1;
         u16 lastItem = entranceScroll + MAX_ITEM_LINES;
         if (lastItem > itemCount) { lastItem = itemCount; }
         Draw_DrawFormattedString(10, 10, COLOR_TITLE, "Randomized Entrances (%d - %d) / %d",
             firstItem, lastItem, itemCount);
 
-        // TODO random entrances visited %
+        // TODO random entrances visited percent?
         // Draw_DrawFormattedString(SCREEN_BOT_WIDTH - 10 - (SPACING_X * 6), 10, itemPercent == 100 ? COLOR_GREEN : COLOR_WHITE, "%5.1f%%", itemPercent);
```
Kewlan: This is a fun idea. Did you forget to add it? If not, this can be removed.

source/entrance.cpp
```diff
@@ -436,6 +442,165 @@ static void ShuffleEntrancePool(std::vector<Entrance*>& entrancePool, std::vecto
...
+  { 0x0008, "Gerudo Fortress Training Grounds Entrance" },
+  { 0x03A8, "Gerudo Training Grounds" },
+  { 0x0433, "Kokiri Forest at Mido's House" },
+  { 0x0443, "Mido's House" },
+  { 0x0437, "Kokiri Forest at Saria's House" },
+  { 0x0447, "Saria's House" },
+  { 0x009C, "Kokiri Forest at House of Twins" },
+  { 0x033C, "House of Twins" },
+  { 0x00C9, "Kokiri Forest at Know-It-All House" },
+  { 0x026A, "Know-It-All House" },
```
Kewlan: I think it'd be better if instead of `at` it said `to` or `from`, like you did further below.

Kewlan (changes requested):

code/src/gfx.c
Giometric: I did forget! Added it in the latest commit.

code/src/gfx.c
```diff
@@ -376,37 +392,37 @@ static void Gfx_DrawSpoilerItemGroups(void) {
...
+
             u32 color = isDiscovered ? COLOR_GREEN : COLOR_WHITE;
-            Draw_DrawFormattedString_Small(10, locPosY, color,
-                "0x%04X -> 0x%04X, BW: 0x%04X", entrOverride.index, entrOverride.override, entrOverride.blueWarp);
+            const char* unknown = "???";
+            const char* startName = pair.StartStrOffset != ENTRANCE_INVALID_STRING_OFFSET ? &gEntranceTrackingData.StringData[pair.StartStrOffset] : "START STRING NOT FOUND";
+            const char* returnName = pair.ReturnStrOffset != ENTRANCE_INVALID_STRING_OFFSET ? &gEntranceTrackingData.StringData[pair.ReturnStrOffset] : "RETURN STRING NOT FOUND";
+
+            Draw_DrawFormattedString_Small(10, locPosY, color, "%s %c", isDiscovered ? startName : unknown, H_DOUBLE_ARROW_CHR);
+            Draw_DrawFormattedString_Small(10, itemPosY, color, "  %s", isDiscovered ? returnName : unknown);
```
Kewlan:
```suggestion
            Draw_DrawFormattedString_Small(10, locPosY, color, "%s %c", gSettingsContext.ingameSpoilers || isDiscovered ? startName : unknown, H_DOUBLE_ARROW_CHR);
            Draw_DrawFormattedString_Small(10, itemPosY, color, "  %s", gSettingsContext.ingameSpoilers || isDiscovered ? returnName : unknown);
```
I forgot about this yesterday.

source/entrance.cpp
```diff
@@ -436,6 +442,165 @@ static void ShuffleEntrancePool(std::vector<Entrance*>& entrancePool, std::vecto
...
+  { 0x01E1, "Spirit Temple" },
+  { 0x0037, "Graveyard Warp Pad" },
+  { 0x0205, "Shadow Temple" },
+  { 0x0098, "Kakariko Bottom of the Well Entrance" },
+  { 0x02A6, "Bottom of the Well" },
+  { 0x0088, "ZF Ice Cavern Entrance" },
+  { 0x03D4, "Ice Cavern" },
+  { 0x0008, "GF Training Grounds Entrance" },
+  { 0x03A8, "Gerudo Training Grounds" },
+  { 0x0433, "KF from Mido's House" },
```
Kewlan, line 468: You don't have to change this, but I feel like it'd make more sense to use `to` instead of `from`, as what it says now implies the entrance that's inside Mido's House that leads to KF. Same with a lot (if not all) other names that say `from`.

Kewlan (changes requested):

source/entrance.cpp
Giometric: I'm half and half on it, 'cause technically it's not wrong (the entrance is "[arriving in] Kokiri forest from Mido's House") but you're right that some of them sound weird written that way. I'm kind of torn between choosing whatever sounds the most straightforward for every entrance, or keeping them all in a semi-consistent format.
Could maybe go the other way, and use `[origin] to [destination]` for all of them?

PR #381, opened 2021-10-08
Add "Fast Bunny Hood" option

Kewlan:
Increases running speed by 50% while wearing the bunny hood.
Does _not_ increase animation speed. The patch might need to be applied somewhere else for that.

### DIFF
--- a/code/include/z3D/z3Dactor.h
+++ b/code/include/z3D/z3Dactor.h
@@ -209,7 +209,9 @@ typedef struct {
 
 typedef struct {
     /* 0x0000 */ Actor  actor;
-    /* 0x01A4 */ char   unk_148[0x00A8];
+    /* 0x01A4 */ char   unk_1A4[0x13];
+    /* 0x01B7 */ u8     currentMask;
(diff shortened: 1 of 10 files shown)


gamestabled:
I want to wait on this and see if we can figure out the animation thing before merging it in, but still cool to see as is

code/Makefile
```diff
@@ -54,7 +54,7 @@ endif
 
 VERFLAGS := -D OOT3D=$(OOT3D) -D OOT3DJ=$(OOT3DJ) -D Z3D=$(Z3D)
 
-ARCH	:=	-march=armv6k -mtune=mpcore -mfloat-abi=softfp -mtp=soft
+ARCH	:=	-march=armv6k -mtune=mpcore -mfloat-abi=softfp -mtp=soft -mfpu=vfpv2
```
lioncash: Would you be able to split this out into its own PR in the meantime? This can be merged, given we should be specifying this anyway

Kewlan: Done

lioncash: Thanks!

PR #385, opened 2021-10-15
Clear rSeqMaxes before setting the max amount

Kewlan:
The unmapped reads from SFX Shuffle would happen when generating multiple seeds without restarting the app, causing `rSeqMaxes` to increase more than it should without resetting, which would then mess up the modulus during gameplay.


Kewlan:
New commit: Removes custom sfx modulo function now that it's fixed with the updated make file

PR #391, opened 2021-10-23
Add entrance tracker groups page

Kewlan:
image

Revamps some of the previous entrance tracking code, like moving `entranceNames` to the patch side and removes the `gEntranceTrackingData` patch.
~~Because the tracker gets the entrance list from `rEntranceOverrides` now, the tracker will include the duplicate entrances between Hyrule Field and Zora's River, but I feel like the effort to remove that pair alone may not be worth it. Both will still count as discovered when entering one of them.~~ Fixed in latest commit.

Will conflict with #386, but the rebase should be simple.

Also always write ingame spoiler log, regardless of what the `Generate Spoiler Log` option is set to.


gamestabled:
Commenting on this here since we talked about it in discord: We are still going to want a way to disable the ingame tracker (probably primarily for races, but also for anybody who dislikes having that available), but that can be a separate change.

PR #396, opened 2021-10-28
2 bug fixes

Roberto-Nessy:
Claim check was considered obtainable from Biggoron when shuffle trade quest was disabled. Fixed by placing an ice trap in that location instead of the claim check.

Logic tricks that were supposed to be always disabled could get enabled. Fixed by commenting them out of the all tricks toggle code.


source/item_pool.cpp
```diff
@@ -744,7 +744,7 @@ void GenerateItemPool() {
     PlaceItemInLocation(DMT_TRADE_BROKEN_SWORD, PRESCRIPTION, false, true);
     PlaceItemInLocation(ZD_TRADE_PRESCRIPTION, EYEBALL_FROG, false, true);
     PlaceItemInLocation(LH_TRADE_FROG, EYEDROPS, false, true);
-    PlaceItemInLocation(DMT_TRADE_EYEDROPS, CLAIM_CHECK, false, true);
+    PlaceItemInLocation(DMT_TRADE_EYEDROPS, ICE_TRAP, false, true);
```
gamestabled: I'm not sure this is the right fix, because if we ever include the option to start with a random adult trade item, with the regular adult trade quest (i think this is how ootr does it) this isn't going to work anymore. I think the root cause of this should be somewhere else in the logic, that made it think we could trade the eyedrops to biggoron, without the pocket egg (or any previous trade item) being in the pool

Roberto-Nessy: I think it's part of all locations reachable that means the location access has to always returns true when the trade quest isn't shuffled. If that feature was added it would be possible to fix it by changing the location access code instead of the item.

gamestabled: Okay, I see. In that case, should we actually just remove this entire block of code (everything in the "else")? It isnt actually doing anything as far as I recall, since locations without a placed override should default to their vanilla item

Roberto-Nessy: Wouldn't that mean the randomiser could try to place items in those locations?

gamestabled: Not if we dont put them in the location pool in those conditions

So i guess, instead of filling the vanilla items into these locations when the trade shuffle is off, we should only include these locations at all when the trade shuffle is on. Does that seem to make the most sense?

PR #402, opened 2021-11-02
Add ingame "Autosave" option

Kewlan:
Will autosave progress when traveling between areas after the set amount of time has passed.
It's set to "Never" by default.

### DIFF
--- a/code/oot.ld
+++ b/code/oot.ld
@@ -692,6 +692,10 @@ SECTIONS
 		*(.patch_SaveGame)
 	}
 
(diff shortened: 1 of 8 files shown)


cjohnson57:
This is really cool! Especially great for Citra, and especially for racing on Citra

Kewlan:
Can at the moment throw fatal errors at what feels like random times:
image

Kewlan:
I may be coming back to this some other time.

PR #403, opened 2021-11-02
Gerudo's Fortress Key Pack setting

cjohnson57:
Just adding to the pile of PRs that won't be added until after 3.0 :Burgy:

Adds an additional setting for Gerudo's Fortress keys, "Anywhere (Pack)", which will allow you to find a single pack of all 4 GF keys. Shamelessly stolen from Freddy's key pack for the treasure chest shop keys.

This makes it so that you can have GF on default openness so that you still get 4 randomized checks, but can also shuffle both card and gerudo keys and have it actually be plausible to get the card check before finishing the seed, also making it much more likely that the card check will actually have something important. Or just if you want to not shuffle card, but do want to have to find the keys to obtain it, but not all 4.

Note if Gerudo's Fortress is set to fast, this setting is no different from normal "Anywhere" since only 1 key will be shuffled in anyway.


cjohnson57:
Closing this so I can redo it with the approach I used over at 64- https://github.com/TestRunnerSRL/OoT-Randomizer/pull/1490

PR #407, opened 2021-11-04
Add Legend view for Dungeon Items page of in-game menu

Giometric:
This PR adds a Legend to the Dungeon Items page, viewable by pressing A while on that page of the in-game menu. The legend can be toggled back off with A, or by navigating to another page of the in-game menu or closing the menu.
image


gymnast86:
Would it be possible to add "-" to the legend as "Neither Barren nor WotH" (or something similar) for when it's used to represent dungeons importance?

Giometric:
Updated layout and added non-WotH / non-barren item
itemLegend_02

PR #408, opened 2021-11-04
Add Grotto Entrances to Entrance Randomizer

gymnast86:
The grottos around Hyrule can now be shuffled around with each other. Grottos work a lot differently from regular entrances which brings about some significant changes to the way entrance randomizer works.

Previously entrance randomizer was handled by swapping around data within the entrance table where necessary. However this approach doesn't work for grotto entrances since a lot of them share the same entrance indices and grotto return points don't have _any_ entries in the entrance table at all. Because of this, entrance randomizer has been reworked to override the next entrance a player goes to on the fly as they trigger a loading zone/cutscene command terminator/grotto entrance.

Since individual grottos loads/return points generally don't have unique entrance table indices, all grotto loads are assigned an entrance index of 0x1000 + a unique grotto ID, and all grotto return points use an index of 0x2000 + a unique grotto ID (this is what OoTR does so I just reused it here). The grotto ID is used to index into either the `grottoLoadTable` or `grottoReturnTable` to retrieve the information necessary for each grotto load/return point.


gamestabled:
I believe there is still an entrance override needed in Entrance_CheckEpona in entrance.c. I wouldve left this comment right there but github won't let me since its not where you had changes i guess

PR #411, opened 2021-11-05
Several fixes

### LINKED ISSUE #384: Spoiler Log bug
Hello,

In my seed in sphere 5 i needed to get the item from the carpet salesman but this isn't displayed in the spoiler log.
reverse wasteland is on. logiclly this is reachable
[1566595914 (Frog, Fairy Ocarina, Deku Stick, Boomerang, Big Magic)-spoilerlog.txt](https://github.com/gamestabled/OoT3D_Randomizer/files/7354212/1566595914.Frog.Fairy.Ocarina.Deku.Stick.Boomerang.Big.Magic.-spoilerlog.txt)
e.

cjohnson57:
Resolves #384

Fixes three issues of varying severity.

1\) Bombchu logic bug
Previously there were 2 issues with the bombchu logic helpers. The first is that with bombchu drops on, it would consider you to have repeatable bombchu access, even though drops are only active if you've found bombchus at least once. The second was that bombchu drops (from Bowling) didn't count for a way to find bombchus. Along with this, I changed the way the bombchu logic helpers are named and laid out to make it more clear what they do.

Edit: Also corrected bombchu purchaseability logic. Before it was either having the bomb bag or bombchus being in logic and having repeatable bombchus, it should only be bombs if bombchus aren't in logic, and if they are you only need to have found bombchus once rather than having repeatable access.

2\) Playthrough generation bug
The playthrough generation part of the search algorithm does preprocessing to remove certain things from the playthrough we know we won't need: Token amounts after the last that might be useful, bombchus after the first, and buy item copies. However, this is in conflict with the way the algorithm for paring down the playthrough works: The playthrough checks if an item is necessary by removing it and seeing if the playthrough is still beatable. For items with multiple copies, such as the bow, this ensures the first 2 bows will be removed and the playthrough will still be beatable, but when the last is removed it no longer will be, so the final bow will be kept in the playthrough. However, for things such as bombchus, since all the other bombchu drops are not added to the playthrough in the first place, removing that first bombchu drop makes it so that the search algorithm will still find other bombchus and thus the game will still be beatable so bombchus would never be printed in the playthrough.

This was resolved by adding a parameter to the search function that tells it to "ignore" certain item locations, by not adding their location if, for example, ignore is set to bombchus and the location contains bombchus. Some might say this sounds like a lot of work and not worth the optimization the preprocessing provides, but those people aren't me. 

3\) Corrected French buy Bomb names


source/logic.cpp
```diff
@@ -518,8 +519,10 @@ namespace Logic {
     Fish         = HasBottle && FishAccess;
     Fairy        = HasBottle && FairyAccess;
 
-    HasBombchus   = (BuyBombchus5 || BuyBombchus10 || BuyBombchus20 || AmmoDrops.Is(AMMODROPS_BOMBCHU)) && FoundBombchus;
-    FoundBombchus = (BombchusInLogic && (Bombchus || Bombchus5 || Bombchus10 || Bombchus20)) || (!BombchusInLogic && BombBag);
+    FoundBombchus   = (BombchuDrop || Bombchus || Bombchus5 || Bombchus10 || Bombchus20);
+    CanPlayBowling  = (BombchusInLogic && FoundBombchus) || (!BombchusInLogic && BombBag);
+    HasBombchus     = (BuyBombchus5 || BuyBombchus10 || BuyBombchus20 || (AmmoDrops.Is(AMMODROPS_BOMBCHU) && FoundBombchus)) && CanPlayBowling;
```
gamestabled: Should this be "or CanPlayBowling" instead of "and CanPlayBowling"?

cjohnson57: (Discussed in Discord) we determined CanPlayBowling actually doesn't need to be there at all, and that the logic for buying bombchus is flawed and also needs correcting.

PR #416, opened 2021-11-12
Hide spoiler location groups and item locations if they are in undiscovered dungeons

### LINKED ISSUE #414: Spoiler log - If dungeon status (vanilla/MQ) is unknown, dungeon sections should be hidden in the spoiler log.
Currently, if a run is generated to not show the dungeon status on the dungeon summary screen, it is still possible to see whether the dungeon is vanilla or Master Quest by scrolling through the spoiler log. If one is looking through the locations to consider what areas to check next, scrolling past the dungeon sections is generally unavoidable.

I assume that if the spoiler log is set to fully show stuff, this is a non-issue. But when it's set to only show the full data for items that have been collected, then this situation applies.

Giometric:
This PR obscures the names of spoiler items located within undiscovered dungeons, and hides the item location groups for those dungeons as well. This fixes an issue where the item tracker easily reveals whether a dungeon is MQ or Vanilla, even if the user has not discovered the dungeon by either finding the map (if enabled in settings) or entering the dungeon at least once. If the user's settings have all dungeons always revealed, no items or groups are hidden.

image
image


code/src/gfx.c
```diff
@@ -142,6 +170,29 @@ static bool IsEntranceDiscovered(s16 index) {
     return isDiscovered;
 }
 
+static bool IsDungeonDiscovered(DungeonId dungeonId) {
+    if (dungeonId <= DUNGEON_GERUDO_TRAINING_GROUNDS) {
+        if (gSettingsContext.dungeonModesKnown) {
+            return true;
+        }
```
cjohnson57, line 177: Is this something that's set if all dungeons are V or MQ, or is it a separate setting?

Giometric: It's set by the app if no dungeon modes will be randomly selected (a mix of MQ and Vanilla can still be all known if the MQs were all added in specifically by the user).

cjohnson57 (approved):

PR #420, opened 2021-11-14
Add functionality to toggle between organizing entrance groups by source or destination

Kewlan:
Either group entrances by where they're from...
image
or by where they lead to.
image

Note how the text by the X button changes. This is the best solution I have with how little space there's left.


code/src/entrance.c
```diff
@@ -111,14 +117,14 @@ void Entrance_Init(void) {
     // Then overwrite the indices which are shuffled
     for (size_t i = 0; i < ENTRANCE_OVERRIDES_MAX_COUNT; i++) {
 
+        if (EntranceIsNull(&rEntranceOverrides[i])) {
+            continue;
```
Kewlan: Now that it checks all values in the struct for 0, this could maybe be `break` instead?

code/src/gfx.c
```diff
@@ -617,17 +622,33 @@ static void Gfx_DrawERTrackerGroups(void) {
...
+            const char* unknown = "???";
+
+            const char* origSrcName = GetEntranceName(destList[locIndex].index);
+            const char* origDstName = GetEntranceName(destList[locIndex].destination);
+            const char* rplcSrcName = gSettingsContext.ingameSpoilers || isDiscovered ? GetEntranceName(destList[locIndex].override) : unknown;
+            const char* rplcDstName = gSettingsContext.ingameSpoilers || isDiscovered ? GetEntranceName(destList[locIndex].overrideDestination) : unknown;
+
+            Draw_DrawFormattedString_Small(10, locPosY, color, "%s to %s %c", origSrcName, origDstName, RIGHT_ARROW_CHR);
+            Draw_DrawFormattedString_Small(10, itemPosY, color, "  %s from %s", rplcDstName, rplcSrcName);
+        }
```
Kewlan: I tried using a struct array pointer, but it didn't work as intended when reading the values so I ended up duplicating the code instead.

Kewlan:
New commit: Entrances are sorted so that they always appear in the following order in the tracker:
- Overworld
- Interior
- Grottos
- Dungeon

Kewlan:
New commit: Differentiates entrances by the previously listed types with different colors. They'll still be white when they're undiscovered. (Don't mind the percent counter, all were colored just for testing.)
image

gamestabled:
The coloring is pretty interesting, do you have plans for how that would work with mixed pools in the future?

Kewlan:
> The coloring is pretty interesting, do you have plans for how that would work with mixed pools in the future?

With the current code it would use the same color for the source and destination, but it would only take two line changes to make the color separate, so it's just up to what we think fits best. 

... but I've added it already as I feel it's pretty obviously better

Issue #426, opened 2021-11-23
Skipping the chest that requires Ruto in MQ Jabu Jabu could potentially soft lock if there's something needed in that chest


Thunderbird91:
In MQ Jabu Jabu, if the player immediately takes Ruto on the path to the Big Octo, there will be nothing left in the main B1 room to hold down the switch to the room on the right. If there's something valuable in that chest, the player will be unable to complete the seed.

A similar issue was present in the vanilla dungeon, but this was fixed with [#413](https://github.com/gamestabled/OoT3D_Randomizer/pull/413).

1. Enter MQ Jabu Jabu with the Boomerang.
2. Clear out the tentacles in the upper area.
3. Descend to the room where Ruto is.
4. Immediately take Ruto to the Big Octo room.

PR #427, opened 2021-11-24
Bug fixes and adjustments

HylianFreddy:
- Malon's song can be obtained without talking to her 3 times beforehand.
- In MQ Jabu, a wooden box appears in the basement after Ruto disappears from the dungeon, so that the Boomerang room can still be accessed afterwards. This resolves issue #426

https://user-images.githubusercontent.com/82058772/143149472-a9856eea-dcb0-4196-9402-6977dde0b3c7.mp4

- Time Blocks had an issue where you could make them disappear while standing on them. I realized the patches I had made for them were overly complicated, so I replaced them with a simpler one and this bug disappeared :)

https://user-images.githubusercontent.com/82058772/143149508-0f7dd515-ff04-4676-85c7-24bf84c8fec0.mp4

- The ambiguous hints were using the wrong character for the apostrophes, which appeared as an asterisk in-game. Now they use the correct one.

### DIFF
--- a/code/src/patches.s
+++ b/code/src/patches.s
@@ -1456,7 +1456,7 @@ TurboTextSignalNPC_patch:
 .section .patch_SkipSongReplays_TimeBlocksFix
 .global SkipSongReplays_TimeBlocksFix_patch
 SkipSongReplays_TimeBlocksFix_patch:
-    moveq r1,#0x1
+    b hook_SkipSongReplays_TimeBlocksFix
 
 .section .patch_ItemsMenuNumSprites
 .global ItemsMenuNumSprites_patch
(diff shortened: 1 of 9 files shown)


code/src/patches.s
```diff
@@ -1453,15 +1453,10 @@ TurboTextClose_patch:
...
-    b hook_SkipSongReplayForTimeBlocksOne
-
-.section .patch_SkipSongReplayForTimeBlocksTwo
-.global SkipSongReplayForTimeBlocksTwo_patch
-SkipSongReplayForTimeBlocksTwo_patch:
-    b hook_SkipSongReplayForTimeBlocksTwo
+.section .patch_SkipSongReplays_TimeBlocksFix
+.global SkipSongReplays_TimeBlocksFix_patch
+SkipSongReplays_TimeBlocksFix_patch:
+    moveq r1,#0x1
```
Kewlan: This should only be replaced if `Skip Song Replays` is enabled, right? Otherwise it changes vanilla behaviour, which some may not like.

HylianFreddy: Thanks, I somehow totally forgot about that.

PR #429, opened 2021-11-25
Development

MemelordRye13:
Could we get x2 speed with bunny hood like im majoras mask?


cjohnson57:
Why did you open this as a pull request lmao

Anyway, if you look at the other pull requests, there is an open one for this feature. But there are issues to work out with it and it won't be until post 3.0.

gamestabled:
@ #381

PR #433, opened 2021-12-01
force the ocarina staff textbox to always close when suns song finishes

gamestabled:
This patch forces the ocarina staff textbox to close at the same that Links stop playing the ocarina when sun's song is played, under any conditions.

This fixes the bug where the textbox stays up for too long when ocarina song shuffle is turned on, but song replays are not skipped.

@HylianFreddy is normally the ocarina songs guy, so Ill let him glance at this before merging


HylianFreddy:
I think this can be merged, but unfortunately it doesn't solve all the problems.
Major issue:
- the MQ Spirit songs puzzle still doesn't work, the gate doesn't open when the song is played.

Minor issues:
- this patch doesn't apply when playing the song for the frogs, but at least Link stays properly locked in the "playing ocarina" state until the music staff disappears.
- it also doesn't apply in places where it usually triggers a scene reload, which means you can still walk around with the text box on screen if you play it for Skull Kid or Darunia.

gamestabled:
Okay, the plan now for 3.0 seems to be to just remove sun's song from the pool for ocarina music shuffle for now. This makes the other patch in the PR redundant for now, but I think we might as well keep it in for a safeguard and for the future.

PR #435, opened 2021-12-02
Perfect the entrance tracker sorting

Kewlan:
Tracker improvements:
- Now sorts the entrances in the order of `entranceData`, after sorting by group and then type. I've manually ordered the entrances to mostly go by what a normal playthrough reaches first.
- `destList` is now also properly sorted so the source and destination of both lists now match.
- Adds a legend page that shows what the four entrance colors are for.

Code improvements:
- Combines the two `draw entrance tracker` functions into one to save space and no longer requires us to duplicate code. I was going to do the same for the `draw item location tracker` functions, but it looked more advanced so I'll wait with that.
- Renames some variables to make more sense.

image


gamestabled:
Sorry for the slow responses to this, irl has taken a lot of priority for me recently. Anyway, this seems good to me except for the 4 chosen colors. I know for a fact that I cant distinguish the 0x00FF00 green and the 0xFF9800 orange thanks to your screenshots, and I suspect that the 0x00A0FF blue and 0xD200FF would also be challenging. I just did a bit of research and here is a suggested palette of 4 colors that nearly anybody should be able to distinguish:

0xE69F00
0x56B4E9
0x009E73
0xF0E442

These are just the first four colors from the Okabe_Ito palette on this page: https://thenode.biologists.com/data-visualization-with-flying-colors/research/ but i think any selection of 4 from that palette should look good if you have a different preference.

Kewlan:
I picked the colors that were the most similar to the old ones:
image

PR #437, opened 2021-12-04
Add submenu with options that set the default ingame options

Kewlan:
image
The options' category are set to `Cosmetic`.


gamestabled:
I think we need to decide which, if any, of these should be considered "Cosmetic" (ie, not affect the seed). I think that Music and SFX should be considered cosmetic, and the rest should not be (the rest all can or do affect gameplay in some way). Does anybody else have thoughts on this?

EDIT: We discussed this in the discord. They will all be cosmetic 👍

PR #440, opened 2021-12-14
Epona Check Fix

gymnast86:
Previously during Overworld ER, the Epona Check would determine if it was okay to keep Link on Epona based on what scene was being loaded. This works most of the time, but created some weird cases (like entering with Epona into Lake Hylia from the Zoras Domain entrance). This check now compares the next entrance index against an array of valid entrances where it's okay to keep Link on Epona if he entered riding her.

### DIFF
--- a/code/src/entrance.c
+++ b/code/src/entrance.c
@@ -346,11 +346,32 @@ u8 EntranceCutscene_ShouldPlay(u8 flag) {
 
 void Entrance_CheckEpona() {
     s32 entrance = gGlobalContext->nextEntranceIndex;
-    s8 dest = gEntranceTable[entrance].scene;
-    //If Link is riding Epona but he's about to enter a scene where she can't spawn,
+    //If Link is riding Epona but he's about to go through an entrance where she can't spawn,
     //unset the Epona flag to avoid Master glitch, and restore temp B.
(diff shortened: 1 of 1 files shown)


code/src/entrance.c
```diff
@@ -346,11 +346,32 @@ u8 EntranceCutscene_ShouldPlay(u8 flag) {
...
     s32 entrance = gGlobalContext->nextEntranceIndex;
-    s8 dest = gEntranceTable[entrance].scene;
-    //If Link is riding Epona but he's about to enter a scene where she can't spawn,
+    //If Link is riding Epona but he's about to go through an entrance where she can't spawn,
     //unset the Epona flag to avoid Master glitch, and restore temp B.
-    if (gSettingsContext.shuffleOverworldEntrances && entrance != 0x496 && (PLAYER->stateFlags1 & 0x00800000) &&
-        dest != 81 && dest != 87 && dest != 90 && dest != 93 && dest != 99) {
+    if (gSettingsContext.shuffleOverworldEntrances && (PLAYER->stateFlags1 & 0x00800000)) {
+
+        static const s16 validEponaEntrances[13] = {
```
Kewlan:
```suggestion
        static const s16 validEponaEntrances[] = {
```
Doing this and using `sizeof(validEponaEntrances) / sizeof(validEponaEntrances[0])` at the for loop will handle the size count automatically.

gymnast86: Right, thanks

PR #462, opened 2022-01-12
Sheik in IC logic fix

Roberto-Nessy:
With closed DoT, starting as child, giant's wallet, a bottle, and dungeon ER it was possible the rando would logically expect you to get Sheik in IC as child even though that's impossible right now.


gamestabled:
It's interesting, this does point out the fact that being able to get the Sheik At Colossus as child is a bit of an inconsistency in-world. I don't think we should change that, especially since it's something everybody has likely come to know by now, but it probably is a bit strange to newcomers.

gamestabled:
Also, it's good and consistent for now, but it also reminds us of the fact that we have a lot of "being adult implies having a sword" logical debt that we need to take care of before attempting any type of master sword shuffle. (the logic isnt the only issue of course, there are technical issues in-game as well)

PR #466, opened 2022-01-17
create conditionally always hints

gamestabled:
I need to test this still, but this way people can see it and tell me if i did something wrong (I havent worked on hints directly before)


gamestabled:
I have now done some testing with this and it seems to be working

PR #469, opened 2022-01-24
Logic trick for flaming chests

gamestabled:
This adds a logic trick to open flaming chests in GTG and Spirit Temple. This is inspired by a particularly nasty async race seed where the hints were effectively useless because they all led to the (otherwise useless) hammer for the GTG check


cjohnson57 (approved):
Looks good. I actually wonder why this isn't a trick in 64

PR #478, opened 2022-02-11
Add local multiplayer

Kewlan:
Uses the UDS service. Temporarily uses a special effect to show where other players are.

**How to use:**
To use, first enable the options in the app. When launching the game, it looks for any existing hosts. If none are found it creates a lobby itself. 
If shared progress is on, whoever is the first to join with a unique seed should load into their savefile first. 
It's possible for two groups of players, each group with their own seed, to see each other while only sharing progress with those with the same seed.

These are the options:
image

Here's a clip that shows the temporary "soul" like effect, and how the web actor is synced between players:

https://user-images.githubusercontent.com/5352197/153521213-564c00ed-d730-4c26-a5d4-083f891c281f.mp4

 **Issues:**
More actors can be synced, but other than that...
- Items can be duplicated if they're collected from a location at the exact same time. Fixing this may take a lot of effort, but at least this isn't something that will make runs unbeatable.
- Citra has some issues: When a player quits the game they still count as connected, so rejoining won't work. There is a temporary workaround though, which is to rejoin the Citra lobby. This will however not reduce the connected player count, so this trick is only possible until the max count is reached.
- Another Citra issue is that it crashes when allowing too many players. The max should be 16, which it is on console, but this has been reduced to 8 on emulator.

Special thanks to the devkitPro team for the easy access to the service functions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *vscode/*
+*vs/*
 backup/*
 build/*
 *.txt
--- a/code/Makefile
+++ b/code/Makefile
@@ -65,7 +65,7 @@ CFLAGS	+=	$(INCLUDE) -DARM11 -D_3DS $(VERFLAGS)
 CXXFLAGS	:= $(CFLAGS) -fno-rtti -fno-exceptions -std=gnu++11
 
 ASFLAGS	+=	-g $(ARCH) $(VERFLAGS)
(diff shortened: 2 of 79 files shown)


MemelordRye13:
Is this citra exclusive? Or can the local multiplayer work on 3ds?

PR #499, opened 2022-04-23
Multiplayer: Fullsync adjustment

Kewlan:
Previously when a player started their game and requested a fullsync, all players with the same seed hash would send their data. I thought this would be fine, but it seems like receiving that many packets that big caused some of them to be discarded.
Now only the first player that starts their game (per seed hash within the network) will send these fullsyncs. That player becomes what I call a "Seed Hash Host".

The negative aspect of this is that if that host leaves, if someone else with the same seed hash wants to join after, the other players will have to restart their games so that someone else becomes a Seed Hash Host, so that new player can get synced. This will probably rarely even happen though, so I think it's fine to change things this way.

### DIFF
--- a/code/src/multiplayer.c
+++ b/code/src/multiplayer.c
@@ -26,6 +26,10 @@ u32 receivedPackets = 0;
 bool duplicateSendProtection = false;
 static s8 netStage = 0;
 static bool mSaveContextInit = false;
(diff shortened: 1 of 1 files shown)


Kewlan:
Turns out this was not enough to solve the problem, but this is how it should have worked from the start anyway, so this can still be merged.

PR #500, opened 2022-04-25
Add ingame "Ignore Mask Reaction" option

Kewlan:
Causes NPCs to respond normally when wearing a mask. Does not apply when trading masks.

https://user-images.githubusercontent.com/5352197/165096330-9d6ea1d5-902a-4a00-a170-6ad315f25f2c.mp4

Also has some code improvements in `gfx_options.c`.

### DIFF
--- a/code/src/hooks.s
+++ b/code/src/hooks.s
@@ -1295,6 +1295,17 @@ hook_SendDroppedBottleContents:
     pop {r0-r12, lr}
     bx lr
 
+.global hook_IgnoreMaskReaction
+hook_IgnoreMaskReaction:
+    ldrh r0,[r0,#0x0]
+    push {r0-r12, lr}
+    cpy r0,r4
+    bl SaveFile_GetIgnoreMaskReactionOption
+    cmp r0,#0x1
+    pop {r0-r12, lr}
+    moveq r0,#0x0
+    b 0x36BBC8
+
 .section .loader
 .global hook_into_loader
 hook_into_loader:
--- a/code/src/savefile.c
+++ b/code/src/savefile.c
@@ -589,6 +589,14 @@ void SaveFile_SetOwnedTradeItemEquipped(void) {
     }
 }
 
+s8 SaveFile_GetIgnoreMaskReactionOption(u32 reactionSet) {
+    // This option somehow breaks talking to the Kakariko Mountain Gate guard, so use a workaround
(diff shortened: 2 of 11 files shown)


code/src/savefile.c
```diff
@@ -589,6 +589,17 @@ void SaveFile_SetOwnedTradeItemEquipped(void) {
...
 }
 
+s8 SaveFile_GetIgnoreMaskReactionOption(void) {
+    // This option somehow breaks talking to the Kakariko Mountain Gate guard, so use a workaround
+    if (PLAYER->currentMask == 1 && gGlobalContext->sceneNum == 0x52 &&
+        gSaveContext.linkAge == AGE_CHILD && (gSaveContext.infTable[7] & 0x80) == 0 &&
+        PLAYER->actor.world.pos.x > -230 && PLAYER->actor.world.pos.x < 170 &&
+        PLAYER->actor.world.pos.z > -1498 && PLAYER->actor.world.pos.z < -1098) {
+        return 0;
+    }
```
HylianFreddy:
```suggestion
s8 SaveFile_GetIgnoreMaskReactionOption(u32 reactionSet) {
    // This option somehow breaks talking to the Kakariko Mountain Gate guard, so use a workaround
    if (reactionSet == 0x3C &&
        PLAYER->currentMask == 1 && (gSaveContext.infTable[7] & 0x80) == 0) {
        return 0;
    }
```
The DMT Gate guard uses reactionSet 0x3C.

code/src/hooks.s
```diff
@@ -1295,6 +1295,16 @@ hook_SendDroppedBottleContents:
     pop {r0-r12, lr}
     bx lr
 
+.global hook_IgnoreMaskReaction
+hook_IgnoreMaskReaction:
+    ldrh r0,[r0,#0x0]
+    push {r0-r12, lr}
+    bl SaveFile_GetIgnoreMaskReactionOption
```
HylianFreddy:
```suggestion
    push {r0-r12, lr}
    cpy r0,r4
    bl SaveFile_GetIgnoreMaskReactionOption
```
The reactionSet is in r4

HylianFreddy (changes requested):
The function you're hooking into receives the reactionSet as a parameter, so you can use that to recognize the DMT guard.

code/src/savefile.c
Kewlan: This is what I tried at first, but I think the reason it didn't work was because I copied `r1` instead of `r4` because of how it was called in Ghidra. Whoops.

PR #510, opened 2022-05-07
Multiplayer: Add Sync ID option

Kewlan:
With this it's possible for two groups of people to play on the same seed hash but only sync their progress with their own group.
For example, this is useful for a 2v2 race if everyone plays on the same network and still wants to be able to see each other.

Also has some code improvements in `multiplayer.c` to reduce repeated code, and make it easier to modify in the future.


Kewlan:
Marking as draft because I want to change the `MP_SyncId` option choices to the `NumericalOptions` function in #511.

Kewlan:
Key loss fix has been included in this PR.

PR #513, opened 2022-05-08
Bug fixes

HylianFreddy:
- Entering Spirit Temple from Desert Colossus clears the temporary flags if the dungeon is randomized.
- Gold Skulltulas check the dungeon mode from the randomizer settings instead of the MQ flag, so that the ones inside blocks in MQ dungeons don't become tangible with mirror world disabled.
- The blue warp patches have been changed to fix the ER destination error when entering a blue warp multiple times. Some patches for the item override inside the Chamber of the Sages have been removed because the cutscenes are skipped now.
- Other minor fixes.

### DIFF
--- a/code/include/z3D/z3D.h
+++ b/code/include/z3D/z3D.h
@@ -525,11 +525,11 @@ typedef void (*DisplayTextbox_proc)(GlobalContext* globalCtx, u16 textId, Actor*
 #define DisplayTextbox_addr 0x367C7C
 #define DisplayTextbox ((DisplayTextbox_proc)DisplayTextbox_addr)
 
-typedef u32 (*EventCheck_proc)(u32 param_1);
+typedef u32 (*EventCheck_proc)(u32 flag);
 #define EventCheck_addr 0x350CF4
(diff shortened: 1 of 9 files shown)


code/include/z3D/z3D.h
```diff
@@ -169,7 +169,8 @@ typedef struct {
     /* 0x156F */ u8           buttonStatus[5];
     /* 0x1574 */ char         unk_1574[0x000F];
     /* 0x1584 */ u16          magicMeterSize;
-    /* 0x1586 */ char         unk_1586[0x000C];
+    /* 0x1586 */ char         unk_1586[0x0004];
+    /* 0x158A */ u16          eventInf[4];
```
Kewlan, line 173: Just asking to make sure since it isn't used anywhere in this PR, was this meant to be added?

HylianFreddy: Kind of. I tested the fix for the temp flag issue by porting over my flag editor from the practice patch, and so I had to add the eventInf array to the SaveContext. When I committed the changes I just decided to leave that in.

PR #517, opened 2022-05-18
Mostly glitch logic (but also a lot else)

Roberto-Nessy:
- Added more glitch logic
  - Added 14 new glitch categories and 14 misc. glitch settings
  - Added overworld glitch logic
  - Added glitch logic for vanilla DT, DC, JJB, FoT, FiT, and WaT
  - Logic for these dungeons has been completely rewritten to a room-by-room format to avoid inferred logic
- Changed logic tricks
  - Enabled some logic tricks marked as needs testing
  - Added some new logic tricks
  - Fixed logicFlamingChests not being set by group toggle
- Changed logic to account for item usability settings
  - Not all situations are accounted for but nothing should cause an unbeatable seed
- Changed health logic to allow starting with lower health
  - Fewer Tunic Requirements now checks health and can apply in more places
  - MQ specific logic has not been altered so may be incompatible with this setting
  - Current health is always updated to not be greater than maximum health
  - The low health beep is disabled when the maximum health is 1 heart
- Changed Closed Forest setting behaviour when using overworld or interior ER
  - Deku Tree entrance is no longer guaranteed to be vanilla
  - Using LW shortcuts and warp songs is in logic before beating DT
- Changed starting bombchus to allow none, 20, or 50
- Changed bottle system to count number of bottles (relevant for some glitches)
  - This also counts Ruto's Letter after it has been delivered
- Changed small key count check
  - Now takes the dungeon KEY instead of the variable tracking the keys as input
- Changed CanUse function
  - Can no longer be used to infer age when checking age-restricted items


cjohnson57:
Great work on this, but a couple comments about some of the creeped-in features

> Changed health logic to allow starting with lower health

Personally I'm not really a fan of this, health being in logic and making heart pieces/containers major items/woth potentially doesn't feel good to me

> Changed Closed Forest setting behaviour when using overworld or interior ER

I'm also not really a fan of this, closed forest is a bad setting but its one reason for existing is to be close to vanilla for beginner rando players, these changes take that reasoning away, imo if we take these things away we may as well just remove closed forest (not that I'd complain)

> Changed CanUse function. Can no longer be used to infer age when checking age-restricted items

What's the reason for this change? Is it for item usability logic? If that's the case couldn't you just do like `is_adult or bow_usability`?

source/setting_descriptions.cpp
```diff
@@ -1420,40 +1416,262 @@ string_view GlitchHoverDescIntermediate               = "Some hovers may require
...
+                                                        "the correct frame.";                              //
+string_view GlitchBombOIDescExpert                    = "You may be expected to use restricted items to\n" //
+                                                        "perform ocarina items where Link's position isn't\n"
+                                                        "particularly lenient";                            //
+/* Hover Boost */                                                                                          //
+string_view GlitchHoverBoostDescDisabled              = "Equipping hover boots when Link takes damage will\n"
+                                                        "cause him to keep the high knockback speed and\n" //
+                                                        "lets him traverse large gaps.\n"                  //
+                                                        "If performed at the edge of a platform Link will\n"
+                                                        "instead perform a super jump which has less range\n"
```
HylianFreddy:
```suggestion
                                                        "instead perform a mega jump which has less range\n"
```
That's usually called Mega Jump

source/setting_descriptions.cpp
```diff
@@ -1420,40 +1416,262 @@ string_view GlitchHoverDescIntermediate               = "Some hovers may require
...
+                                                        "the water during the catching cutscene.";         //
+string_view GlitchCutsceneDiveDescAdvanced            = "You may be expected to use magic arrows to perform"
+                                                        "Farore's Wind cutscene dives.";                   //
+/* Navi Dive: Stick */                                                                                     //
+string_view GlitchNaviDive_StickDescDisabled          = "A Navi dive is a type of cutscene dive achieved by"
+                                                        "falling off a ledge while talking to Navi. While\n"
+                                                        "the usual method for achieving this is a TSC, it\n"
+                                                        "is also possible to perform using a jump attack\n"//
+                                                        "with deku sticks.";                               //
+string_view GlitchNaviDive_StickDescNovice            = "You may be expected tp enter BotW with a stick\n" //
```
HylianFreddy:
```suggestion
string_view GlitchNaviDive_StickDescNovice            = "You may be expected to enter BotW with a stick\n" //
```
Typo

Roberto-Nessy:
I personally see WotH hearts as similar to WotH GS tokens - still minor items but possibly necessary. Through testing I never saw more than 6 hearts added to the playthrough and none of them were ever WotH as they could be replaced with tunics or other heart containers/pieces or Nayru's love.

The closed forest behaviour only changes with overworld or interior ER enabled, without them it stays the same as before. This was done since it's almost impossible to not escape the forest with those ER settings so the option doesn't really make sense anymore, and also if you're using those settings you're probably not a beginner rando player anyway.

The CanUse function was changed as part of the item usability logic so `CanUse(ADULT_ITEM)` returns `HasItem && (IsAdult || UsabilitySetting)`. I brought it up because a decent amount of the logic was written using CanUse as an implicit age check, eg: `GV_CHEST` previously only checked `CanUse(MEGATON_HAMMER)` whereas now it explicitly needs the `IsAdult` check.

source/location_access.cpp
```diff
@@ -741,17 +822,24 @@ void AreaTable_Init() {
...
+                  LocationAccess(GV_CHEST,          {[]{return IsAdult && (CanUse(MEGATON_HAMMER) || LogicGVHammerChest);},
+                                         /*Glitched*/[]{return IsAdult && (CanDoGlitch(GlitchType::LedgeCancel, GlitchDifficulty::NOVICE) || CanDoGlitch(GlitchType::BombHover, GlitchDifficulty::INTERMEDIATE) || (CanUse(STICKS) && CanDoGlitch(GlitchType::QPA, GlitchDifficulty::NOVICE) && (CanTakeDamageTwice || CanDoGlitch(GlitchType::QPA, GlitchDifficulty::ADVANCED))));}}),
                   LocationAccess(GV_TRADE_SAW,      {[]{return IsAdult && PoachersSaw;}}),
-                  LocationAccess(GV_GS_BEHIND_TENT, {[]{return CanUse(HOOKSHOT) && AtNight && CanGetNightTimeGS;}}),
-                  LocationAccess(GV_GS_PILLAR,      {[]{return CanUse(HOOKSHOT) && AtNight && CanGetNightTimeGS;}}),
+                  LocationAccess(GV_GS_BEHIND_TENT, {[]{return IsAdult && CanUse(HOOKSHOT) && AtNight && CanGetNightTimeGS;},
+                                         /*Glitched*/[]{return (CanDoGlitch(GlitchType::BombHover, GlitchDifficulty::INTERMEDIATE) || CanUse(BOOMERANG)) && IsAdult && AtNight && CanGetNightTimeGS;}}),
```
Kewlan: Speaking of Gerudo Valley. Should `CanUse(BOOMERANG)` maybe go into the glitchless func above? (Same for any similar instances.)

Also regarding opening grottos with hammer, maybe that should be a Glitch Trick so it's not just blindly assumed, and is more clear to anyone willing to try this Glitched logic?

HylianFreddy (changes requested):
Cool stuff.
There's a couple of minor mistakes in the setting descriptions

source/location_access.cpp
Roberto-Nessy: Ah yeah, that's part of what I meant when I said 
> Not all situations are accounted for but nothing should cause an unbeatable seed

I do plan to go back through at some point and make sure all the item usability changes are considered everywhere but my initial pass was just making sure none of it inferred an age check. (That specific line is left over from when item usability was going to be only for glitch logic, but there are other similar cases)

Also I don't remember adding hammer to any grotto entrances that weren't already using CanBlastOrSmash, though I could be wrong.

Kewlan: Is was thinking about how, for example, `KAKARIKO_VILLAGE` to `KAK_REDEAD_GROTTO` checks `CanUse(MEGATON_HAMMER)` in the Glitched function, but upon looking closer I see `CanUse(MEGATON_HAMMER)` is already checked in `CanBlastOrSmash`, so I suppose this counts as what you were going to look over again. So all good. 👍

source/logic.cpp
```diff
@@ -541,40 +729,47 @@ namespace Logic {
...
+    CanTakeDamage       = Fairy || CanSurviveDamage;
+    CanTakeDamageTwice  = (Fairy && NumBottles >= 2) || ((EffectiveHealth == 2) && (CanUse(NAYRUS_LOVE) || Fairy)) || (EffectiveHealth > 2);
     //CanPlantBean        = IsChild && (MagicBean || MagicBeanPack);
     CanOpenBombGrotto   = CanBlastOrSmash       && (ShardOfAgony || LogicGrottosWithoutAgony);
     CanOpenStormGrotto  = CanPlay(SongOfStorms) && (ShardOfAgony || LogicGrottosWithoutAgony);
     HookshotOrBoomerang = CanUse(HOOKSHOT) || CanUse(BOOMERANG);
-    CanGetNightTimeGS = (CanPlay(SunsSong) || !NightGSExpectSuns);
+    CanGetNightTimeGS   = (CanPlay(SunsSong) || !NightGSExpectSuns);
+
+    Health          = BaseHealth + HeartContainer + (PieceOfHeart >> 2);
```
Kewlan: When I see `Health` I imagine the exact value. Maybe it should be called `Hearts` instead?
```suggestion
    Hearts          = BaseHearts + HeartContainer + (PieceOfHeart >> 2);
```

source/logic.cpp
```diff
@@ -541,40 +729,47 @@ namespace Logic {
...
+    EffectiveHealth = ((Health << (2 + DoubleDefense)) >> Multiplier) + ((Health << (2 + DoubleDefense)) % (1 << Multiplier) > 0); //Number of half heart hits to die, ranges from 1 to 160
+    FireTimer       = CanUse(GORON_TUNIC) ? 255 : Health;
+    WaterTimer      = CanUse( ZORA_TUNIC) ? 255 : Health;
```
Kewlan: For clarity I think the timers should say in seconds:
```suggestion
    FireTimer       = CanUse(GORON_TUNIC) ? 255 : (Hearts * 8);
    WaterTimer      = CanUse( ZORA_TUNIC) ? 255 : (Hearts * 8);
```

source/location_access.cpp
```diff
@@ -1562,79 +1733,92 @@ void AreaTable_Init() {
...
   areaTable[DMC_UPPER_LOCAL] = Area("DMC Upper Local", "Death Mountain Crater", DEATH_MOUNTAIN_CRATER, NO_DAY_NIGHT_CYCLE, {
                   //Events
-                  EventAccess(&GossipStoneFairy, {[]{return GossipStoneFairy || (HasExplosives && CanSummonGossipFairyWithoutSuns);}}),
+                  EventAccess(&GossipStoneFairy, {[]{return GossipStoneFairy || (HasExplosives && CanSummonGossipFairyWithoutSuns && FireTimer > 2);}}),
                 }, {
                   //Locations
-                  LocationAccess(DMC_WALL_FREESTANDING_POH, {[]{return true;}}),
-                  LocationAccess(DMC_GS_CRATE,              {[]{return IsChild && CanChildAttack;}}),
-                  LocationAccess(DMC_GOSSIP_STONE,          {[]{return HasExplosives;}}),
+                  LocationAccess(DMC_WALL_FREESTANDING_POH, {[]{return FireTimer > 1;}}),
```
Kewlan: All of these that only have `FireTimer` still need `CanUse(GORON_TUNIC) || LogicFewerTunicRequirements` since players can appear from them once mixed and decoupled entrance shuffling is added, right? 

And a small nitpick: I think all timers should be standardized to `>=`.

Kewlan (commented):
Took a look at the timer stuff.

Kewlan:
Since there are so many changes, maybe people would prefer if this was split up and discussed per feature. 

We could start by splitting up `location_access.cpp` and adding the glitches (outside of location_access, so logic and settings etc.), and then a PR could be submitted per area/location_access file, and per logic change, like the new areas and the heart stuff. 

While this would require a lot of rewriting, I think it may be beneficial. It'd also be an opportunity to clean up the code, ~like with the `CanUse` changes that are already needed~.

source/location_access.cpp
Roberto-Nessy: I think the logic in this area was already written with that in mind so continuing to not check FTR should be fine, however I have added that the player needs at least 3 hearts if not using FTR (and also moved all the FTR checks inside the Timer variables)

PR #518, opened 2022-05-18
Patch Side: Use sub-directories

Kewlan:
image

GitHub currently ignores file renames when looking at file history. Go here to see the moved files before they were moved: https://github.com/gamestabled/OoT3D_Randomizer/tree/49854d6198a882c9c19711491762dd7c6ca66668/code/src


Kewlan:
Turns out you can also use `Blame` to see the history!
(Note the folder structure at the top left, while past changes are still on the left.)
image

PR #524, opened 2022-06-24
Bug fixes and minor improvements

HylianFreddy:
This PR fixes some issues related to the Boss Challenge:
- multiplayer mode now pauses during BC to avoid overwriting the inventory erroneously
- all bosses are available from the start, to avoid confusion due to their vanilla unlock requirement being stones/medallions owned.
- the boss intro cutscenes are removed (as they should've been in the vanilla game)

Some other bugs have been fixed:
- the skulltula at the top of MQ Fire is now forced to have the proper rotation
- Saria's hints work correctly again (they broke when the "Responding to your mask" text was removed)

I have also included the following changes and minor tweaks:
- Truth Spinner rotation speed is increased
- Target Deku Scrub in Lost Woods gives reward faster
- Restored Link's House as a savewarp location, because I don't think it has any negative effects.
- A new in-game option to change the "Skip Song Replays" setting
- The Shooting Gallery text reminder for the quiver requirement stays on screen longer without interrupting movement.


Kewlan (changes requested):
Pausing the receiving of packets will cause them to queue up. Over a hundred packets could be received per second with multiple players. I don't know how the service handles this, but I think it's safe to assume that old packets will be tossed after a while.

So instead of pausing multiplayer, pause shared progress instead:
- Return here if in boss challenge: https://github.com/gamestabled/OoT3D_Randomizer/blob/main/code/src/multiplayer.c#L518
- Skip these two scopes if in boss challenge:
https://github.com/gamestabled/OoT3D_Randomizer/blob/main/code/src/multiplayer.c#L2522
https://github.com/gamestabled/OoT3D_Randomizer/blob/main/code/src/multiplayer.c#L2541
- That might be enough, if not then this function needs to be called after exiting the challenge:
https://github.com/gamestabled/OoT3D_Randomizer/blob/main/code/src/multiplayer.c#L203

You can see it'll work similar as when the player is in the main/file select menu, so "in the back" progress will still be synced.

HylianFreddy:
@Kewlan I can do that, but isn't the problem you mentioned (packets queuing up) already happening whenever a player is waiting on the title screen?

Or does this also take care of that? https://github.com/gamestabled/OoT3D_Randomizer/blob/main/code/src/gfx.c#L1055

Kewlan:
Yeah exactly. 
`Multiplayer_Update` is called in three different places so it constantly gets called.
One is during normal gameplay, one is while the custom info menu is open, and that one is while in the main menu.

HylianFreddy:
Ok, then I think I can just move the check to `IsInGame`

code/src/gfx.c
```diff
@@ -1052,7 +1052,7 @@ void Gfx_Update(void) {
         Gfx_ShowMultiplayerSyncMenu();
     }
 
-    // The update is called here so it works while in file select
+    // The update is called here so it works while in different game modes (title screen, file select, boss challenge, credits, MQ unlock)
```
Kewlan, line 1055: The `Multiplayer_Update` below here is no longer called while in the boss select menu, which can cause syncs to fail.

Kewlan (changes requested):
I found one issue. Otherwise it looks fine, but it'd be nice if we could keep the ability to see others while in the boss challenge. As I mentioned, only shared progress needs to be paused.

code/src/gfx.c
HylianFreddy: Thanks, I had misplaced the cutscene check in IsInGame, oops.

I changed the check to send ghost data but I haven't tested it yet. I'll do it later if you don't do it in the mean time.

[Ok I'm testing now and it's crashing lol. I'll see how to fix it]

HylianFreddy: I think everything works now. Let me know if there are still issues

Kewlan (changes requested):
Seems good, just one more thing: Link sfx aren't sent while in boss challenges, so this `IsInGame` needs to be changed to `IsInGameOrBossChallenge`:
image

HylianFreddy:
> Seems good, just one more thing: Link sfx aren't sent while in boss challenges, so this `IsInGame` needs to be changed to `IsInGameOrBossChallenge`: image

Done!

Kewlan:
- Does `Skip Song Replays` work even when `Quick Text` is off? If yes and since it's possible to have it enabled now when `Quick Text` is off, should `Skip Song Replays`'s visibility no longer be controlled by `Quick Text`?
- I don't think the Shooting Gallery Man needs to say the reminder text every single time you win, so how about making him only say it once per play session? (In other words by using a local variable)

HylianFreddy:
> - Does `Skip Song Replays` work even when `Quick Text` is off? If yes and since it's possible to have it enabled now when `Quick Text` is off, should `Skip Song Replays`'s visibility no longer be controlled by `Quick Text`?

Yes, patch-wise those two things are unrelated.
I originally created the dependency to reduce clutter in the menu if someone wanted to play with vanilla text, but I realize probably nobody does that anyway. I'll remove it if you think it's better.

> - I don't think the Shooting Gallery Man needs to say the reminder text every single time you win, so how about making him only say it once per play session? (In other words by using a local variable)

I assume you mean a static variable (so a "play session" ends when you close the software, not when you go back to the title screen). That's easy to do, but personally I think the message should stay.
I purposefully made the text behave that way to not cause any interruption, so I don't think it would bother anyone who wants to play the unrandomized mini-game.
I can also imagine some of the "unaware" players might even forget about the bow requirement in the same play session (maybe they use sleep mode instead of closing the game).

Kewlan:
Alright, the reminder message can stay as it is. 👍

And yeah I think `Skip Song Replays` should always be visible now. (And without the indentation)

PR #534, opened 2022-08-06
Initial work for shuffling frog song rupees

garrettjoecox:
We're in the process of adding this into SoH ([PR](https://github.com/HarbourMasters/Shipwright/pull/1061)) and figured we'd provide you guys with a starting point if you want it! 

I do not have a building environment or testing environment set up for this project so this might not even build in it's current state. 

This does not include edits to the actual patches, but it should be as simple as handling the other `FrogSongType`'s from `code/include/z3D/actors/z_en_fr.h` in `code/src/actors/frog.c`

If this is not helpful feel free to close it out!


gamestabled:
@HylianFreddy has already actually worked on the rest of making this functional on a branch of his, so I will merge this, then merge his, and it will all be good to go. Thanks!

PR #537, opened 2022-08-18
Make ISG patch optional

HylianFreddy:
The default option will keep ISG enabled.
Disabling it could be interesting to make more challenging playthroughs with no logic or glitched logic (after all why would you do super stabs if you can just do ISG and keep your stick), or just to stay closer to vanilla OoT3D.


gamestabled:
been meaning to make this an option forever. thanks for making it

PR #538, opened 2022-08-18
Implement arrow quick-swapping

HylianFreddy:
This PR adds the ability to swap arrow type while aiming the bow (like in Wind Waker), and a setting to choose the button to do so.

https://user-images.githubusercontent.com/82058772/185499784-b2cdbff3-d8cd-4fd4-ae73-cb71e728ebb2.mp4

### DIFF
--- a/code/include/input.h
+++ b/code/include/input.h
@@ -6,6 +6,10 @@ typedef struct {
     btn_t up;
     btn_t pressed;
     btn_t old;
+    int16_t touchX;
+    int16_t touchY;
+    uint32_t touchPressed;
+    uint32_t touchHeld;
 } InputContext;
 
 void Input_Update(void);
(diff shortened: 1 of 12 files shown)


Kewlan:
Looks like it's only possible to advance the arrow type. Could you make it so it's possible to go back too? For example, when on Normal Arrows and you want to go to Light Arrows you just press the designated "back" button once instead of forward thrice. To reflect this in the options you could change it to `D-Pad Right/Left` and `D-Pad Up/Down`.

Also I'm curious: Does swapping drain magic, or is that fixed with the `magicTarget` and `magicState` stuff you added?

HylianFreddy:
> Also I'm curious: Does swapping drain magic, or is that fixed with the `magicTarget` and `magicState` stuff you added?

It's fixed, swapping out an arrow restores the magic that was consumed to nock it.

> Looks like it's only possible to advance the arrow type. Could you make it so it's possible to go back too? For example, when on Normal Arrows and you want to go to Light Arrows you just press the designated "back" button once instead of forward thrice. To reflect this in the options you could change it to `D-Pad Right/Left` and `D-Pad Up/Down`.

Yes, but I feel like that's a bit unnecessary. Consider these things:
- You don't often have all arrow types while playing
- Some players already use button remapping, and using up more buttons could complicate things for them
- You also only have one button to cycle through the arrows in Wind Waker and other rom-hacks like MM3D Restoration or Ship of Harkinian

gamestabled:
beautiful

PR #548, opened 2022-09-24
Master Sword Shuffle, Age restrictions logic toggle & Critical health patches

Roberto-Nessy:
- Fully functional master sword shuffle
- Adds toggle for considering items used as the wrong age in logic
- Adds patches to prevent critical health effects when max health is below 2 hearts


code/src/savefile.h
```diff
@@ -24,6 +24,7 @@ void SaveFile_InitExtSaveData(u32 fileBaseIndex);
 void SaveFile_LoadExtSaveData(u32 saveNumber);
 void SaveFile_SaveExtSaveData(u32 saveNumber);
 void SaveFile_EnforceHealthLimit(void);
+u8 SaveFile_SwordlessPatchesEnabled(void);
 
 // Increment the version number whenever the ExtSaveData structure is changed
 #define EXTSAVEDATA_VERSION 10
```
Kewlan: This should be updated

code/oot.ld
```diff
@@ -1374,4 +1374,52 @@ SECTIONS
...
+        *(.patch_CriticalHealthCheckOne)
+    }
+
+    .patch_CriticalHealthCheckTwo 0x4716D4 : {
+        *(.patch_CriticalHealthCheckTwo)
+    }
+
+    .patch_CriticalHealthCheckThree 0x49F294 : {
+        *(.patch_CriticalHealthCheckThree)
+    }
```
HylianFreddy: These should be ordered by address to be consistent with the rest of the file.
(and the `0x44fd98` should be capitalized)

code/src/savefile.c
```diff
@@ -686,3 +691,43 @@ void SaveFile_EnforceHealthLimit(void) {
...
+        }
+    }
+
+    // If first time going adult update adult equips (equip master sword if in inventory)
+    if (gSaveContext.adultEquips.equipment == 0) {
+        gSaveContext.adultEquips.equipment = 0x1120 | (gSaveContext.equipment & 0x2);
+        if (gSaveContext.equipment & 0x2) {
+            gSaveContext.adultEquips.buttonItems[0] = 0x3C;
+        }
+    }
```
HylianFreddy, line 719: I think this is going to patch out adult reset. Going forward in time with 0xFF on adult B should trigger the code in this `if` block.

HylianFreddy (changes requested):
Cool stuff.
I suggest adding a setting for Master Sword to the starting inventory too. It would also be nice to not see the sword on Link's back if you don't have it, but that would probably require digging through the draw functions, so maybe it's not worth it.

code/src/savefile.c
Roberto-Nessy: It will run the original code if the master sword has already been found on the file and the B value is 0xFF (SaveFile_SwordlessPatchesEnabled() becomes false and return 0 goes to original code for becoming adult with 0xFF). I've added the line that sets equipment to the first condition to simulate adult reset before obtaining the master sword

Roberto-Nessy:
> It would also be nice to not see the sword on Link's back if you don't have it, but that would probably require digging through the draw functions, so maybe it's not worth it.

Yeah that and making the master sword model not appear during the Ganon fight at all are things I'd like to make happen eventually but aren't necessary for it to be functional right now.

Kewlan:
Can you add a sync exception for Master Sword in multiplayer when in the final Ganon fight scene? Here: https://github.com/gamestabled/OoT3D_Randomizer/blob/92477761d5f68047050d56cc38bab8bb66c46c03/code/src/multiplayer.c#L592-L607
Also about `masterSwordFlags`, do you think they need to be synced? Depending on when they're set maybe they can be fit into existing syncing functions?

Roberto-Nessy:
1 of the master sword flags gets set when receiving the master sword item and the other when the pedestal item is pushed so they do probably need syncing. They never get unset so if there are any sync functions that just or together the values that would be sufficient.

The way the bits change in the Ganon fight shouldn't be any different than it was before so should be fine unless it already needed handling.

PR #560, opened 2022-10-07
Ability to specifically set King Zora's speed

Kestron06:
Add the ability to set King Bon III's shuffle count specifically.

### DIFF
--- a/code/src/actors/king_zora.c
+++ b/code/src/actors/king_zora.c
@@ -20,8 +20,8 @@ void EnKz_rUpdate(Actor* thisx, GlobalContext* globalCtx) {
     // Zora's Domain scene, the position check is just to add a 1 frame delay
     if (gSettingsContext.kingZoraSpeed != KINGZORASPEED_VANILLA && globalCtx->sceneNum == 88 && thisx->speedXZ != 0 && thisx->world.pos.x < 628.0f) {
         u8 mweepCount = (gSettingsContext.kingZoraSpeed)? Bias(Hash(0)) + 1 : 1;
-        if(gSettingsContext.kingZoraSpeed==KINGZORASPEED_CUSTOM){
-            mweepCount=gSettingsContext.exactZoraSpeed+1;
+        if (gSettingsContext.kingZoraSpeed == KINGZORASPEED_CUSTOM){
+            mweepCount = gSettingsContext.exactZoraSpeed + 1;
         }
         thisx->speedXZ = 6.4f / (2.56f * (mweepCount - 0.58f));
         thisx->world.pos.z = -1783.0f; // lock Z position so the increased speed doesn't mess it up
--- a/code/src/actors/king_zora.c
+++ b/code/src/actors/king_zora.c
@@ -20,6 +20,9 @@ void EnKz_rUpdate(Actor* thisx, GlobalContext* globalCtx) {
     // Zora's Domain scene, the position check is just to add a 1 frame delay
     if (gSettingsContext.kingZoraSpeed != KINGZORASPEED_VANILLA && globalCtx->sceneNum == 88 && thisx->speedXZ != 0 && thisx->world.pos.x < 628.0f) {
         u8 mweepCount = (gSettingsContext.kingZoraSpeed)? Bias(Hash(0)) + 1 : 1;
+        if(gSettingsContext.kingZoraSpeed==KINGZORASPEED_CUSTOM){
+            mweepCount=gSettingsContext.exactZoraSpeed+1;
+        }
         thisx->speedXZ = 6.4f / (2.56f * (mweepCount - 0.58f));
         thisx->world.pos.z = -1783.0f; // lock Z position so the increased speed doesn't mess it up
     }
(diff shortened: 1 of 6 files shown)


source/settings.cpp
```diff
@@ -265,7 +265,8 @@ namespace Settings {
...
   Option BigPoeTargetCount   = Option::U8  ("Big Poe Target Count",   {NumOpts(1, 10)},                                                       {bigPoeTargetCountDesc});
   Option NumRequiredCuccos   = Option::U8  ("Cuccos to return",       {NumOpts(0, 7)},                                                        {numRequiredCuccosDesc});
-  Option KingZoraSpeed       = Option::U8  ("King Zora Speed",        {"Fast", "Vanilla", "Random"},                                          {kingZoraSpeedFast, kingZoraSpeedVanilla, kingZoraSpeedRandom});
+  Option KingZoraSpeed       = Option::U8  ("King Zora Speed",        {"Fast", "Vanilla", "Random", "Custom"},                                {kingZoraSpeedFast, kingZoraSpeedVanilla, kingZoraSpeedRandom, "Manually set how many shuffles King Zora takes."});
+  Option ExactZoraSpeed      = Option::U8  ("  Exact Shuffle Count",  {NumOpts(1,128)},                                                      {""});
```
Kewlan: - Descriptions go into `descriptions.cpp`. Add the new one here: https://github.com/gamestabled/OoT3D_Randomizer/blob/c24bca840e46786bc2ae4032a0eb2683fd2bb6f7/source/descriptions.cpp#L613-L621
- Rename the option for clarity, since there's nothing shuffled here
- Make sure the spacing isn't off
```suggestion
  Option KingZoraSpeed       = Option::U8  ("King Zora Speed",        {"Fast", "Vanilla", "Random", "Custom"},                                {kingZoraSpeedFast, kingZoraSpeedVanilla, kingZoraSpeedRandom, kingZoraSpeedCustom});
  Option ExactZoraSpeed      = Option::U8  ("  Exact Mweep Count",    {NumOpts(1,128)},                                                       {""});
```

Kewlan (changes requested):
When I read about this being added in discord I was gonna complain about option bloat, but I think you did this in the best, discreet way possible. Well done!

source/settings.cpp
Kestron06: Alright, I saw the whole descriptions.cpp thing I just don't get the point - where I come from we are usually competing with each other for less lines of code while still being legible. I'll go ahead and change that I guess.

Which option? exactZoraSpeed?

It wasn't off in my editor, is there somewhere specifically that I missed?

Kestron06: Also, to make those changes do I need to close the Pull Request?

Kewlan: > Alright, I saw the whole descriptions.cpp thing I just don't get the point - where I come from we are usually competing with each other for less lines of code while still being legible. I'll go ahead and change that I guess.

Well, just imagine how the settings file would look if all descriptions were in it directly. It's also good for in case someone else eventually wanted to add another option after yours, then they don't have to clean up after you

> Which option? exactZoraSpeed?

Yeah

> It wasn't off in my editor, is there somewhere specifically that I missed?

The little description part was off by one, but it's good now

Kestron06: So all good or is there something I missed?

Kewlan: The option is still called `Exact Shuffle Count`, or do you think that makes more sense?

Kestron06: To me that makes sense, especially since the option is indented and only visible after setting King Zora Speed to Custom. But it's not hard to change, what would you recommend? I can't think of anything else.

Kestron06: OH You're talking about changing it to Mweep. Yeah, to be honest I definitely think shuffle count makes more sense. I always hear King Zora saying "Aweek" LOL.

code/src/actors/king_zora.c
```diff
@@ -20,6 +20,9 @@ void EnKz_rUpdate(Actor* thisx, GlobalContext* globalCtx) {
     // Zora's Domain scene, the position check is just to add a 1 frame delay
     if (gSettingsContext.kingZoraSpeed != KINGZORASPEED_VANILLA && globalCtx->sceneNum == 88 && thisx->speedXZ != 0 && thisx->world.pos.x < 628.0f) {
         u8 mweepCount = (gSettingsContext.kingZoraSpeed)? Bias(Hash(0)) + 1 : 1;
+        if(gSettingsContext.kingZoraSpeed==KINGZORASPEED_CUSTOM){
+            mweepCount=gSettingsContext.exactZoraSpeed+1;
+        }
```
gymnast86:
```suggestion
        if (gSettingsContext.kingZoraSpeed == KINGZORASPEED_CUSTOM){
            mweepCount = gSettingsContext.exactZoraSpeed + 1;
        }
```
Could you change the spacing around the operators on these lines to match the surrounding style? Styling usually isn't too important to me, but for some reason this being different from the rest of the file really bothers me.

Kestron06: Sure I guess, I honestly like this format better but yeah I can do that.

Kestron06: Done

PR #566, opened 2022-10-12
Add One-way entrances, Mixed Pools, and Decoupled Pools to Entrance Shuffle

gymnast86:
It may have taken almost a year, but I finally got around to finishing this PR. It adds the following options to Entrance Shuffling:
 - Overworld Spawns
 - Warp Songs
 - Owl Drops
 - Mixed Entrance Pools
 - Decoupled Entrances

The in-game entrance tracker has also been updated to support one-way and decoupled entrances.


code/include/draw.h
```diff
@@ -53,6 +53,7 @@
 #define COLOR_GREEN         RGB8(0x00, 0x9E, 0x73)
 #define COLOR_BLUE          RGB8(0x56, 0xB4, 0xE9)
 #define COLOR_ORANGE        RGB8(0xE6, 0x9F, 0x00)
+#define COLOR_YELLOW        RGB8(0xFF, 0xFF, 0x00)
```
Kewlan: This should be adjusted for colorblindness, see acceptable colors here: https://github.com/gamestabled/OoT3D_Randomizer/pull/435#issuecomment-992089545
The closest, which is still very yellow, would be:
```suggestion
#define COLOR_YELLOW        RGB8(0xF0, 0xE4, 0x42)
```

Kewlan (commented):
(Solved)

The one-way group list is missing the spawns, but the group entrance count is still correct so that puts all entrances in the subsequent lists back two steps:
image

Kewlan:
Did a bit of testing with various settings, running around and looking at the tracker, and it all seems to work great!

While you're at it, what do you think about renaming `Src` and `Dest` in the tracker, to `From` and `To`? It'd read like "From Group" and "To Group".
image

Also, how come decoupled entrances not revealing their override works properly without this needing adjustments? 
(At `// Set connected`) 
https://github.com/gamestabled/OoT3D_Randomizer/blob/6d2ebc3fddc2395b7bb16fe38bc459aa882cf5fd/code/src/savefile.c#L284-L308

Kewlan:
Can you also add all the new options to the Full Chaos preset here: 
https://github.com/gamestabled/OoT3D_Randomizer/blob/6d2ebc3fddc2395b7bb16fe38bc459aa882cf5fd/source/preset.cpp#L375-L384

source/entrance.cpp
```diff
@@ -855,10 +1205,18 @@ void CreateEntranceOverrides() {
...
+
+    s16 destinationIndex = -1;
+    s16 replacementDestinationIndex = -1;
+
+    // Only set destination indices for two way entrances and when decouple entrances
+    // is off
+    if (entrance->GetReverse() != nullptr && !Settings::DecoupleEntrances) {
+      replacementDestinationIndex = entrance->GetReplacement()->GetReverse()->GetIndex();
+      destinationIndex = entrance->GetReverse()->GetIndex();
+    }
```
gymnast86, line 1219: I agree with changing `Src` and `Dest` to `From` and `To`.

`SaveFile_SetEntranceDiscovered` doesn't need to be modified because decoupled entrances have their `destination` and `overrideDestination` fields set to -1. So when the function gets called a second time for the `overrideDestination` it just returns without changing anything.

gymnast86:
I agree with changing `Src` and `Dest` to `From` and `To`.

`SaveFile_SetEntranceDiscovered` doesn't need to be modified because decoupled entrance overrides have their `destination` and `overrideDestination` fields set to -1. So when the function gets called a second time, it just returns after the first if statement.

https://github.com/gymnast86/OoT3D_Randomizer/blob/a326a6fc5887e755a98a19f55a5e2fc28f44c0b9/source/entrance.cpp#L1207-L1219

PR #567, opened 2022-10-13
Force child as starting age with intended DoT and unshuffled Ocarinas

### LINKED ISSUE #565: Seed generation failure with DoT closed (intended) and vanilla ocarinas as adult in glitchless logic
Hello from the Ship of Harkinian! 👋 

While investigating [this issue on our side](https://github.com/HarbourMasters/Shipwright/issues/1747), a conversation in our Discord ensued, with one of our members being able to confirm that, with the options on the title, 3DS rando fails to generate a seed since it seems that Link can't get the Ocarina of Time and thus becomes unbeatable.

Pending a fix on our side, I'll make an upstream PR based on SoH's solution.

gymnast86:
Logic fix for the issue brought up in #565

### DIFF
--- a/source/descriptions.cpp
+++ b/source/descriptions.cpp
@@ -170,9 +170,8 @@ string_view ageDesc                   = "Choose which age Link will start as.\n"
                                         "Forest.\n"                                        //
                                         "\n"                                               //
                                         "Child will also be forced if Door of Time is set\n"
-                                        "to intended, ocarinas are unshuffled, and logic is"
-                                        "set to glitchless since adult Link can't obtain\n"//
-                                        "the Ocarina from the moat.";                      //
+                                        "to intended and ocarinas are unshuffled unless you"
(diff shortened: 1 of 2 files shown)


Kewlan:
What glitch allows getting the last ocarina? For Glitch Logic you get to choose which glitches you should be expected to do and at what difficulty, so you could for example only set ISG at Novice and that's it. So this should maybe not account for the logic setting at all.

Instead it can check if starting ocarinas is at least one.

gymnast86:
Ah true, with current glitched logic it wouldn't be possible. There is a glitch called [Death Hole Wrong Warping](https://www.zeldaspeedruns.com/oot3d/ww/death-hole-wrong-warp) which can be used to skip the Door of Time, but glitched logic doesn't consider it because I imagine it would be a giant headache to try and come up with proper logic for.

PR #589, opened 2022-11-08
Prevent duped decoupled entrances in spoiler log

Archez:
After the change to have decoupled entrances in the spoiler log, it seems some entrances were in there twice.

Adding an `!IsAddedToPool()` for the reverse check similar to the top level check prevents duplicate records.

The only thing I can't fully comment on is if some of the decoupled entrances should be considered in a different "sphere". I'm not sure how the spheres are determined.


Archez:
Updated with the latest, and added back a previous fix that went missing from the clang-format merge.

PR #590, opened 2022-11-09
Complete german translation

Schicksal88:
This randomizer gave me a lot of enjoyment over the last month.
English is fine and all, but playing in your own language is still the best, so I decided to translate it into german myself.

Went over all the files multiple times and played multiple seeds. I couldn't find any typos/bugs so I call it finished.
Thanks for making the Randomizer!

*A note about item location hints:
Most english location hints talk about the activity which grants the item, for example "beat 2 zombies guarding".
I had to change those to refer to the location of the item instead of the activity, so "behind 2 zombies" or "next to 2 zombies" for the german sentences to make sense. If the german hint would "beat 2 zombies", the article of the item would change from "ein" Deku-Stick to "einen" Deku-Stick and then the "item is at location" hint would be incorrect as they need "ein" Deku-Stick.
This leads to some location hints being a little different than the english one. I tried to stay close to the english meaning, but sometimes had to put in something completely different.


gamestabled:
Very awesome, thanks a ton for doing this!

PR #591, opened 2022-11-10
Add option to use Rupees as ammo

CharadeNut:
This is inspired by ALttP Randomizer's Retro game option, where the player uses rupees instead of arrows.
With this option, the player will use ammunition as normal, but once the ammo count is about to reach 0, rupees will be spent to bring it back to 1.
This applies to the bow and slingshot, but also bombs, bombchu, sticks, nuts and magic (magic's threshold is the cost to use Nayru's Love, to make sure every spell can be used with rupees).
Rupee costs are as follow:
• Bow: 4 rupees
• Slingshot: 2 rupees
• Bombs: 8 rupees
• Bombchu: 12 rupees
• Deku Nuts: 4 rupees
• Deku Stick: 15 rupees
• Magic: ~24 rupees to use Nayru's Love, 1 rupee for a charged spin attack


Kewlan:
What do you think about allowing ammo to go to 0, and then charge the rupees on the next item use? Would feel nicer and make more sense, I think.

CharadeNut:
> What do you think about allowing ammo to go to 0, and then charge the rupees on the next item use? Would feel nicer and make more sense, I think.

While I agree that it would be less "hacky", there are a few obstacles to this:
• Either you'd have to find hooks for the various items, so that it directly resorts to rupees rather than decreasing ammo,
• or you'd have to find hooks to give ammo only when various buttons are pressed (which still begs the question of whether the game will correctly assume that the player has enough ammunition)

Either of those cases requires in-depth research and programming that I'm unable to do. If this is something you feel confident to delve into yourself, you've more than welcome to do so.
For all intents and purposes, to me this patch works exactly as intended, is relatively lightweight, and it affects all possible ammo uses.

PR #604, opened 2022-12-01
Fixes and tweak 4

Kewlan:
- Make random cosmetics consistent using a separate generator.
In other words, generating the same seed with the same cosmetic settings multiple times generates the same results every time.
- Remove patch for brown boulders that was intended for multiplayer.
It caused all boulders with the same flag to get destroyed at the same time which was something I didn't consider, like the ones in MQ Dodongo's Cavern.
- Update the item text of Ruto's pickup message in Jabu.
image
It looks ugly but I decided to add a newline before the item name so all of them can fit, like the infamous GTG Small Key.

### DIFF
--- a/source/custom_messages.cpp
+++ b/source/custom_messages.cpp
@@ -981,61 +981,37 @@ void CreateAlwaysIncludedMessages() {
 
     // Ruto picking up the dungeon reward
     {
-        std::string itemColor;
+        u8 itemColor;
         switch (Location(BARINADE)->GetPlacedItem().GetHintKey()) {
             case KOKIRI_EMERALD:
             case FOREST_MEDALLION:
-                itemColor = COLOR(QM_GREEN);
+                itemColor = QM_GREEN;
                 break;
             case ZORA_SAPPHIRE:
             case WATER_MEDALLION:
-                itemColor = COLOR(QM_BLUE);
+                itemColor = QM_BLUE;
                 break;
             case SHADOW_MEDALLION:
-                itemColor = COLOR(QM_PINK);
+                itemColor = QM_PINK;
                 break;
             case LIGHT_MEDALLION:
             case SPIRIT_MEDALLION:
             case GOLD_SKULLTULA_TOKEN:
-                itemColor = COLOR(QM_YELLOW);
+                itemColor = QM_YELLOW;
                 break;
             default:
-                itemColor = COLOR(QM_RED);
+                itemColor = QM_RED;
                 break;
         }
-        CreateMessage(0x4050, 0, 2, 3,
-                      UNSKIPPABLE() + INSTANT_TEXT_ON() + "Princess Ruto got the " + NEWLINE() + itemColor +
-                          Location(BARINADE)->GetPlacedItemName().GetNAEnglish() + COLOR(QM_WHITE) + "!" +
-                          WAIT_FOR_INPUT() + INSTANT_TEXT_OFF() + "But why Princess Ruto?" + MESSAGE_END(),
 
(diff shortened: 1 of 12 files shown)


source/custom_messages.cpp
```diff
@@ -978,6 +978,65 @@ void CreateAlwaysIncludedMessages() {
...
+                          itemColor + Location(BARINADE)->GetPlacedItemName().GetEURSpanish() + COLOR(QM_WHITE) + "!" +
+                          WAIT_FOR_INPUT() + INSTANT_TEXT_OFF() + "Tú te quedas con las ganas..." + MESSAGE_END(),
+
+                      UNSKIPPABLE() + INSTANT_TEXT_ON() + "La Principessa Ruto ha recuperato" + NEWLINE() + "la " +
+                          itemColor + Location(BARINADE)->GetPlacedItemName().GetEURItalian() + COLOR(QM_WHITE) +
+                          "..." + INSTANT_TEXT_OFF() + MESSAGE_END(),
+
+                      UNSKIPPABLE() + INSTANT_TEXT_ON() + "Prinzessin Ruto hat den" + NEWLINE() + itemColor +
+                          Location(BARINADE)->GetPlacedItemName().GetEURGerman() + COLOR(QM_WHITE) + " wieder!" +
+                          WAIT_FOR_INPUT() + INSTANT_TEXT_OFF() + "Aber warum Prinzessin Ruto?" + MESSAGE_END());
```
HylianFreddy:
```suggestion
        u8 itemColor;
        switch (Location(BARINADE)->GetPlacedItem().GetHintKey()) {
            case KOKIRI_EMERALD:
            case FOREST_MEDALLION:
                itemColor = QM_GREEN;
                break;
            case ZORA_SAPPHIRE:
            case WATER_MEDALLION:
                itemColor = QM_BLUE;
                break;
            case SHADOW_MEDALLION:
                itemColor = QM_PINK;
                break;
            case LIGHT_MEDALLION:
            case SPIRIT_MEDALLION:
            case GOLD_SKULLTULA_TOKEN:
                itemColor = QM_YELLOW;
                break;
            default:
                itemColor = QM_RED;
                break;
        }

        Text rutoDialog = Text{ "Princess Ruto got #", "La princesse Ruto a trouvé #", "¡La princesa Ruto tiene #",
                                "La Principessa Ruto ha recuperato #", "Prinzessin Ruto hat #" } +
                          Location(BARINADE)->GetPlacedItem().GetHint().GetClear() +
                          Text{ "#!^But why Princess Ruto?", "#!^Mais pourquoi la princesse Ruto?",
                                "#!^Tú te quedas con las ganas...", "#...", "#!^Aber warum Prinzessin Ruto?" };

        rutoDialog.Replace("$", ""); // Plural marker
        CreateMessageFromTextObject(0x4050, 0, 2, 3, AddColorsAndFormat(rutoDialog, { itemColor }));
```

HylianFreddy (changes requested):
Cool idea to change the Ruto text! There are some issues with the translations though. I'll leave my suggested implementation in a comment.

- Depending on the random item, grammatical gender and number can be off for the article in all languages except English. The hint system already handles those so it should be used here too.
- I think the NA Spanish version references the feminine stone with the word "ella" ("you're left without **it**"). I think it's just better to use the European version, and at that point the 5-languages constructor is good enough.

Current:
Bad French

Suggested:
Good French

source/custom_messages.cpp
Kewlan: Wow, this is awesome. Thank you!

Issue #528, opened 2022-07-13
Game crashes on startup


Lynquid:
Once I open the game I get a black screen, my cursor does the loading animation with the hourglass right next to it and then Citra just closes. This doesn't happen without the randomizer and it never happened with an earlier version.
Also I got two different logs (not two for one crash but two different ones for two separate crashes)
[2 citra_log.txt](https://github.com/gamestabled/OoT3D_Randomizer/files/9102310/2.citra_log.txt)
[1 citra_log.txt](https://github.com/gamestabled/OoT3D_Randomizer/files/9102311/1.citra_log.txt)

Lynquid:
It works with the US version, seems like older versions of the rando worked with EU but not the newest one ^^

PR #606, opened 2022-12-11
Documentation: Improve readability of region restriction

jessestricker:
I've overlooked the statement that this only works with the USA version and spent 30 minutes fiddling with Citra settings, trying different Citra versions and looking in the open and closed issues of this repo.

This change makes the documentation more visible.

__Related:__
- #528 this person also did not see the note

_(edits: grammar)_


Roberto-Nessy:
This is actually kind of outdated as the nightly releases have had EUR version support for over 2 months, although the stable release doesn't have it yet

jessestricker:
The documentation links to the stable versions, so I thought this would be an improvement until a new version is released.

gymnast86:
Yeah, until we have a major release which supports EUR and USA, I think this documentation would be better to emphasize.

PR #607, opened 2022-12-15
Guarantee Goron Pot Reward on First Try

Kewlan:
Only applies when the reward isn't already dropped by the pot, and the reward hasn't been picked up before.


cjohnson57:
Why does this need to be a thing? Timing the pot is pretty easy

gymnast86:
I think this if fine to try out for now since it seems like other players were also in favor of this. If we ever need to revert it/make it an option (i.e. if we ever implement freestanding rupees shuffle) then we can come back to it.

PR #613, opened 2022-12-29
Added Bunny Hood as a starting item option

CharadeNut:


gymnast86:
Would you be able to add a description for the option that says something like the following?
`
This setting is experimental. Note that when you obtain the Weird Egg it will overwrite the Bunny Hood in your inventory and you'll have to open up the mask shop to get it again.
`

This way people who read the description won't be surprised when it happens to them.

gymnast86:
Actually this is simple enough that I can just add it myself

PR #614, opened 2022-12-29
Improve build process

HylianFreddy:
This PR edits the build scripts to introduce the following changes:
- Running `make` in the repo folder will rebuild the basecode patches and repack the 3DSX application.
This allows to rebuild the app relatively quickly after editing files in the `code` folder or the assets archive, without having to run `make clean` beforehand and then wait for all the app's source files to be rebuilt for nothing.
When the previous behavior is desired, the `app_only` flag can be passed on the command line to ignore any changes in the `code` folder (anyway, this would only save the 2 seconds required to check that there are indeed no changes).
- The Python script won't rewrite the HPP symbol files if they are already correct, to improve the app's build speed (mainly for when the basecode files don't change).

Here's a badly cropped video with some random examples (the second build spent 20 seconds on the cpp files instead of the 40 it would've spent after a `make clean`):

https://user-images.githubusercontent.com/82058772/210015604-5aaf3c9c-d7e5-4cf9-a8a5-41ad3c2b56f9.mp4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,9 @@ ifneq ($(debug), 0)
 	CXXFLAGS += -g -DENABLE_DEBUG
 endif
 
+# Enable this to skip building the basecode patches
+app_only ?= 0
+
 #---------------------------------------------------------------------------------
 # list of directories containing libraries, this must be the top level containing
 # include and lib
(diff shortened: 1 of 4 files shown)


Kewlan:
This is amazing. Thank you so much for this!
I tested it and I no longer have to wait 2 minutes every time I updated the patch side. Just a few seconds! ❤

PR #616, opened 2022-12-30
Add clang-format check

Kewlan:
Similar to the other action, but this one's only for the clang-format check.

If it fails, one can click on `Details` and it'll show what files are wrong and how it'll looked when fixed:
image
image

Something more to consider is to make it so pull requests can only be merged when this check succeeds. 
To do so, go to `Settings` -> `Branches` -> `Add branch protection rule`
(Has to be done after this is merged)

### DIFF
--- a/.github/workflows/clang_format_check.yml
+++ b/.github/workflows/clang_format_check.yml
@@ -1,0 +1,16 @@
+name: clang-format Check
+
+on:
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


HylianFreddy (approved):
I agree with adding the branch protection rule.

PhlexPlexico:
This looks good! As an aside, I think you should maybe update actions to use v3, as Node.js 12 support is being deprecated, which will eventually break this action.
image

PR #619, opened 2023-01-02
Increase speed of Shadow Temple boat

Kewlan:
Makes it five times faster.
The cutscene that plays is also shortened.


HylianFreddy (changes requested):
Could you please add the patches to `oot_e.ld` too?
We wouldn't want to give a disadvantage to European players 😄

Kewlan:
Whoops! I forgot!

PR #621, opened 2023-01-04
Add build check

Kewlan:
This apparently works! 
Adds a check that tests if the pull request can be built successfully.

Same as clang check, if it fails click on `Details` and see what's wrong:
image
image

Requesting review from @PhlexPlexico.


PhlexPlexico:
LGTM 👍 :)

.github/workflows/build_check.yml
```diff
@@ -0,0 +1,18 @@
+name: Build Check
+
+on:
+  pull_request:
+
+jobs:
+  build:
```
PhlexPlexico: One nitpick though - maybe just under `build` you could add name like `name: Compile Project` (or whatever you prefer, just something a bit more descriptive?) so it just doesn't have the single `build` when you click into the action?

image

Kewlan: Added names to steps, and combined the checks into one file!

PhlexPlexico (approved):
🔥

PR #624, opened 2023-01-06
Workaround for location tracker ending too early

### LINKED ISSUE #609: Check tracker isn't displaying the last couple of area pages
I've generated multiple seeds on the latest development branch, 3.1-aaf14e, and every time, there are issues with the last couple of area pages of the check tracker concerning Hyrule Castle and Ganon's Castle.

Sometimes, these pages don't show up at all, leaving the Market as the last visible area page on the check tracker. Other times, the Hyrule Castle page shows up, but with an incomplete list of the checks there, while Ganon's Castle still isn't visible.

I've been generating these seeds with all the shuffle item settings on to have the max possible number of checks, and doing so previously has led to the check tracker showing at least 500 checks. But with this issue, only about 480 or so checks are listed. Completing a check that isn't visible doesn't make it show up.

Kewlan:
Puts the item locations and string data in a separate array of patches.
To increase the amount of writable locations, only `SPOILER_LOCDATS` needs to be increased. I tested making it over two thousand locations, which works:

https://user-images.githubusercontent.com/5352197/211091546-4e5ca8ae-218b-4566-882c-34de2925e4df.mp4

Closes #609.

Also fixes some things for the dungeon tracker, and improves some related code.


HylianFreddy:
Good job on the fix!

I wonder if it's possible to do the same for the entrance tracker too, because I tried generating a max settings rando and the entrance list was broken:

image

Here's the seed:
[broken entrance tracker.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/10365955/broken.entrance.tracker.zip)

Kewlan:
Above is fixed by #627.

PR #628, opened 2023-01-07
Allow app and ingame menu to be navigated with the Circle Pad

### LINKED ISSUE #410: Menu control with (circle pad)
version : Nightly-8c7203
Model: New 3ds NSTC-J
Game version: NSTC 1.0

I configured Rehid with OOT to be able to use my ZR,ZL button for touch items and open the item/equipment/map with the dpad.
But remapping the dpad break navigation in the spoiler list.

A possible workaround would be to allow us to navigate the spoiler list using the circlepad instead of restricting it to the dpad only.

Also, pressing select to open the spoiler list would also bring up the save menu after closing the list, it is now fixed but I quite liked that "feature", it was useful when start was remapped to touch as well. Would it be possible to bring it back as a toggle in the options menu?

Thank you for your time

### LINKED ISSUE #570: [Controls Mapping] We should be able to scroll with circle pad as well as d-pad
So, you know how we currently scroll only with d-pad, it'd also make it easier for players who have difficulty tryna get to the dpad to use circle pad, like have an option for us to switch between dpad and circle pad, or at least make it so we can scroll with both instead. It helps very much.

Kewlan:
Closes #410.
Closes #570.

### DIFF
--- a/code/include/hid.h
+++ b/code/include/hid.h
@@ -111,5 +111,9 @@ typedef struct {
 #define BUTTON_L1 (1 << 9)
 #define BUTTON_X (1 << 10)
 #define BUTTON_Y (1 << 11)
+#define CPAD_RIGHT (1 << 28)
+#define CPAD_LEFT (1 << 29)
+#define CPAD_UP (1 << 30)
+#define CPAD_DOWN (1 << 31)
 
 #endif // HID_H
--- a/code/src/gfx.c
+++ b/code/src/gfx.c
@@ -854,7 +854,7 @@ static void Gfx_ShowMenu(void) {
         } else if (curMenuIdx == PAGE_SPHERES && gSpoilerData.SphereCount > 0) {
             // Spoiler log
             u16 itemCount = gSpoilerData.Spheres[currentSphere].ItemCount;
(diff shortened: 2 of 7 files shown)


HylianFreddy:
Unfortunately using the circle pad inside the in-game menu still doesn't seem to work on 3DS, even though it does on Citra.
Navigating the app instead works fine on both platforms.

Kewlan:
Aah, dang. I'll unlink the issues then.

HylianFreddy:
I think I may have a solution.
In `input.c` this block
```
u32 GetCurrentPadState(void) {
    u32 hid_shared_mem = *(u32*)(0x005AEC5C);
    return *(volatile u32*)(hid_shared_mem + 0x1C);
}
#define HID_PAD (GetCurrentPadState())
```
should be replaced with this
```
#define HID_PAD (real_hid.pad.pads[real_hid.pad.index].curr.val)
```

It should be tested to make sure nothing else breaks, but it lets the circle pad work in the menu.

Kewlan:
Awesome, thank you! Worked for me on Citra like normal too 👍

PR #633, opened 2023-01-08
Add "Permadeath" Option

Kewlan:
On death, instantly deletes your save file and kicks you back to the main menu.

There are some safety measures in place. For permadeath to be active, both the seed and save file need to have been created while the option was enabled.

https://user-images.githubusercontent.com/5352197/211195948-c199750c-bedc-4bfd-943f-3bdec97f6956.mp4


gamestabled:
this is horrible and hilarious lmao, nice. did you happen to test how it interacts with multiplayer save file syncing?

Kewlan:
> did you happen to test how it interacts with multiplayer save file syncing?

I didn't, but what should happen is basically nothing. It will still delete the save file (gSaveContext), but the copy that shared progress uses (mSaveContext) will stay intact and will apply on any save load as usual.

PR #634, opened 2023-01-12
Fix list of valid Epona entrances

Roberto-Nessy:
- ZR -> HF was actually HF -> ZR
- LW -> HF was actually ZR -> HF
- GF -> Thieves Hideout is not a valid Epona entrance

### DIFF
--- a/code/src/entrance.c
+++ b/code/src/entrance.c
@@ -494,7 +494,7 @@ void Entrance_CheckEpona(void) {
     // unset the Epona flag to avoid Master glitch, and restore temp B.
     if (gSettingsContext.shuffleOverworldEntrances && (PLAYER->stateFlags1 & 0x00800000)) {
 
-        static const s16 validEponaEntrances[] = {
+        s16 validEponaEntrances[] = {
             0x0102, // Hyrule Field -> Lake Hylia
             0x0189, // Lake Hylia -> Hyrule Field
             0x0309, // LH Fishing Hole -> LH Fishing Island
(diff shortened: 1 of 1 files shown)


code/src/entrance.c
```diff
@@ -504,12 +504,11 @@ void Entrance_CheckEpona(void) {
...
             0x01F9, // Lon Lon Ranch -> Hyrule Field
             0x01FD, // Market Entrance -> Hyrule Field
-            0x00EA, // ZR Front -> Hyrule Field
-            0x0181, // LW Bridge -> Hyrule Field
+            0x0181, // ZR Front -> Hyrule Field
+            0x0185, // LW Bridge -> Hyrule Field
             0x0129, // GV Fortress Side -> Gerudo Fortress
             0x022D, // Gerudo Fortress -> GV Fortress Side
             0x03D0, // GV Carpenter Tent -> GV Fortress Side
-            0x0496, // Gerudo Fortress -> Thieves Hideout
```
HylianFreddy, line 512: This was added intentionally so that the glitch could still be triggered on purpose like in the vanilla game, so I think it should stay.

Original PR for reference: #406

Roberto-Nessy: I guess that's fine at the moment, I was thinking about future proofing in case the hideout entrances get added to ER someday. Maybe it should have some signifier that it needs to be handled if that happens?

HylianFreddy: In that case you can wrap the index in a call to `Entrance_GetOverride`:
```
0x03D0, // GV Carpenter Tent -> GV Fortress Side
Entrance_GetOverride(0x0496), // Gerudo Fortress -> Thieves Hideout
0x042F, // LLR Stables -> Lon Lon Ranch
```

Roberto-Nessy: That works, although it can't use a function result in a static const array so I chose to handle that check separately.

PR #635, opened 2023-01-14
Fixes for MQ icon, savewarp and time of day

HylianFreddy:
- Drawing the icon for Ganon's Castle dungeon mode was reading one of the "skipped trials" settings as the dungeon mode, due to an incorrect index.
- Savewarping in Ganondorf's boss room was going back to the overworld spawn, now it will go to Ganon's Tower.
- Beating dungeons wasn't setting time of day.

### DIFF
--- a/code/src/gfx.c
+++ b/code/src/gfx.c
@@ -449,7 +449,8 @@ static void Gfx_DrawDungeonItems(void) {
         // Dungeon Type
         if (dungeonId <= DUNGEON_GERUDO_TRAINING_GROUNDS || dungeonId == DUNGEON_INSIDE_GANONS_CASTLE) {
             if (IsDungeonDiscovered(dungeonId)) {
-                bool isMasterQuest         = gSettingsContext.dungeonModes[dungeonId] == DUNGEONMODE_MQ;
+                u32 dungeonModeId          = (dungeonId == DUNGEON_INSIDE_GANONS_CASTLE ? 10 : dungeonId);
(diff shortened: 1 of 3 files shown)


code/src/gfx.c
```diff
@@ -449,7 +449,8 @@ static void Gfx_DrawDungeonItems(void) {
         // Dungeon Type
         if (dungeonId <= DUNGEON_GERUDO_TRAINING_GROUNDS || dungeonId == DUNGEON_INSIDE_GANONS_CASTLE) {
             if (IsDungeonDiscovered(dungeonId)) {
-                bool isMasterQuest         = gSettingsContext.dungeonModes[dungeonId] == DUNGEONMODE_MQ;
+                u32 dungeonModeId          = (dungeonId == DUNGEON_INSIDE_GANONS_CASTLE ? 10 : dungeonId);
+                bool isMasterQuest         = (gSettingsContext.dungeonModes[dungeonModeId]) == DUNGEONMODE_MQ;
```
Kewlan, line 453: Thanks, this was something I missed! I'm pretty sure it's wrong for GTG too. I solved this in another function like this: https://github.com/gamestabled/OoT3D_Randomizer/blob/33ecc084f4be1a0ada8dff9a7aa6360d0b6d3abe/code/src/gfx.c#L201-L202 I suggest to globalize this and rename it to something like `idToModeIndex`.

HylianFreddy: I thought of that, but unfortunately it wouldn't work because `dungeonModesKnown` and `dungeonModes` are mapped differently (and the GTG icon works fine currently).
I didn't change them to avoid the risk of screwing up somewhere and cause more bugs.

PR #644, opened 2023-02-07
Add Chest Texture options

HylianFreddy:
This PR replaces the Chest Size option with Chest Appearance, with the same options found in OoTR.

Two new custom textures are added to the assets archive. These textures can definitely be improved, so if someone wants to create better ones I will replace them. The general idea is to have a gold chest for major items, maybe with some added symbols like OoTR does, and a silver version of the boss key chest for small keys. We could also have special chests for Skulltula Tokens and Heart Pieces, if more textures are made for them.

Currently, the chests look like this:
yellow-chest
key-chest

Some notes for creating the custom textures:
The vanilla textures can easily be dumped with Citra. They are sized 128x128 and are almost completely transparent, because they use the alpha channel to represent the metallic "shininess": the higher the alpha value for a pixel, the more environmental light will be reflected off that pixel.

3DSX build:
[OoT3D_Randomizer.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/10679711/OoT3D_Randomizer.zip)


gymnast86:
As mentioned in the discord, this is looking pretty awesome! For the vanilla description could we maybe use something like `Chests will appear as they do in the base game`?

HylianFreddy:
> As mentioned in the discord, this is looking pretty awesome! For the vanilla description could we maybe use something like `Chests will appear as they do in the base game`?

Sure, I'll do that now.

PR #646, opened 2023-02-14
Add "Hyper Bosses" Option

Kewlan:
Makes all bosses, including Ganon, act twice as fast!

### DIFF
--- a/code/oot.ld
+++ b/code/oot.ld
@@ -1380,6 +1380,10 @@ SECTIONS
 		*(.patch_ChildDontEquipSwordSlotByDefault)
 	}
 
+	.patch_HyperActors 0x461790 : {
+		*(.patch_HyperActors)
+	}
+
 	.patch_TitleCardUpdate 0x4618B8 : {
 		*(.patch_TitleCardUpdate)
 	}
@@ -1528,6 +1532,10 @@ SECTIONS
 
 	.patch_DampeSetCollectibleFlag 0x4D9D50 : {
 		*(.patch_DampeSetCollectibleFlag)
+	}
+
+	.patch_PhantomGanonLightningID 0x523698 : {
+		*(.patch_PhantomGanonLightningID)
 	}
 
 	.patch_BombchuCheapestPriceOne 0x525348 : {
--- a/code/oot_e.ld
+++ b/code/oot_e.ld
@@ -1380,6 +1380,10 @@ SECTIONS
 		*(.patch_ChildDontEquipSwordSlotByDefault)
 	}
 
+	.patch_HyperActors 0x4617B0 : {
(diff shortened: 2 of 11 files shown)


Kewlan:
I'm putting this on hold because I want to test doing this by adding a patch in `Actor_UpdateAll` instead. Would make it shorter and more efficient, and easy to add for other actors like mini bosses if desired.

Kewlan:
I tried the other attempt by re-running the `do { ... } while (i < 0xC)` part of `Actor_UpdateAll` for bosses and enemies, but it had issues. See code and described issues in commit message here: https://github.com/gamestabled/OoT3D_Randomizer/compare/main...Kewlan:OoT3D_Randomizer:hyper-bosses-2

This PR's version I've played two 100% seeds already and have noticed no issues. So while it does have a bunch of extra, repeated code, at least it works.

HylianFreddy:
> I tried the other attempt by re-running the `do { ... } while (i < 0xC)` part of `Actor_UpdateAll` for bosses and enemies, but it had issues. See code and described issues in commit message here: [main...Kewlan:OoT3D_Randomizer:hyper-bosses-2](https://github.com/gamestabled/OoT3D_Randomizer/compare/main...Kewlan:OoT3D_Randomizer:hyper-bosses-2)
> 
> This PR's version I've played two 100% seeds already and have noticed no issues. So while it does have a bunch of extra, repeated code, at least it works.

I made an attempt at implementing this feature using only a single patch, on the call to the actual `update` function inside `Actor_UpdateAll`. It seems to work decently, but I haven't added any of the "special cases to update in order" that are in this PR.
If you want to check it out here's my branch: ~~https://github.com/HylianFreddy/OoT3D_Randomizer/tree/hyper-bosses-2~~ [Branch deleted, the commit is [here](https://github.com/HylianFreddy/OoT3D_Randomizer/commit/4f4f3c19bb53f923e99522dac199898e9476a59e)]

Kewlan:
> I made an attempt at implementing this feature using only a single patch, on the call to the actual `update` function inside `Actor_UpdateAll`. It seems to work decently, but I haven't added any of the "special cases to update in order" that are in this PR. If you want to check it out here's my branch: https://github.com/HylianFreddy/OoT3D_Randomizer/tree/hyper-bosses-2

Bongo Bongo and Barinade doesn't work properly like this because their actors need to be updated in normal order.

I'm motivated to keep trying now, so I'll mark as draft again.

Kewlan:
I tried methods that wouldn't require these special cases, but with no success. To use a patch instead is still a great idea though, so I went with that. I haven't added a hyper enemies option here because that will require a lot of testing for other special cases.

PR #652, opened 2023-02-26
Tiny fixes

### LINKED ISSUE #650: Dungeon Item legend has a typo regarding the Boss Key and Map.
Relatively minor but I think it's worth mentioning. As of the current dev version, v3.1-47791b, there seems to be an issue with the dungeon item legend, where the "Boss Key" icon and text are missing, and the "Map" icon is listed as "Maps Key". Perhaps the string "Boss Key" got partially overwritten with "Map", leading to "Maps Key"?
image

Kewlan:
- Fixes Y offset for "Boss Key" text in Dungeon Item Legends
- Fixes typos in Custom Music descriptions

Closes #650


HylianFreddy:
There is also another typo to fix in the two descriptions for the "Own" music setting ("it's").

PR #653, opened 2023-03-11
Extra curse traps

Roberto-Nessy:
Adds 2 new curse traps that cause the camera to rotate
These have a new toggle in advanced traps as they could cause motion sickness
Also fixes a minor issue where some traps would sometimes deal the wrong amount of damage if mirror world used a setting other than on/off


HylianFreddy:
I think `ScreenTraps` should be a sub-setting of `CurseTraps`, given that they're still a type of curse.

PR #656, opened 2023-03-20
Assuming \n works here, this fixes a formatting nuisance

Kestron06:
I don't know whether or not \n works here or not I am making an assumption that it does. If it does not, let me know and reject the pull request.


Kewlan:
Notice how there's no newline between `and` and `save`? That's because hints get auto-formatted. 
A solution that would cover similar cases would be to adjust the max line length here: https://github.com/gamestabled/OoT3D_Randomizer/blob/0fda31366c3ff6d5a8e5fd1e9bf0c60dc88339f0/source/custom_messages.cpp#L652

Kestron06:
Cool, thanks

PR #657, opened 2023-03-21
Free cam

Roberto-Nessy:
Adds a setting to use the C-stick to control the camera like in MM3D
Also adds a setting to the ingame menu to invert the axes

Big thanks to UraYukimitsu for getting C-stick input working


code/src/gfx_options.c
```diff
@@ -66,6 +67,23 @@ void InitOptions(void) {
...
+        strcpy(options[opNum].alternatives[altNum++], "Invert X-Axis");
+        strcpy(options[opNum].alternatives[altNum++], "Invert Both");
+        strcpy(options[opNum].description,
+               "Change the input directions of the free camera.\n\nThis feature is not available on old 3DS systems.");
+    } else {
+        strcpy(options[opNum].alternatives[altNum++], "Disabled");
+        strcpy(options[opNum].description, "Free camera is disabled, enable it in\n\"Misc Settings\".\nThis feature is "
+                                           "not available on old 3DS systems.");
+    }
+    options[opNum++].optionPointer = &gExtSaveData.option_FreeCamControl;
```
Kewlan: You can wrap the entire option in the if statement so it doesn't show up when Free Cam is off. Empty/unused options are already checked for, and ignored.

Kewlan (commented):
I see you decided to add this as an option while also disabling it when racing is on. I think only one is necessary, or what are your thoughts on it?

Roberto-Nessy:
> I see you decided to add this as an option while also disabling it when racing is on. I think only one is necessary, or what are your thoughts on it?

Yeah you're probably right, I think I'll leave just the option. Also thanks for mentioning the gfx_options thing, I probably should've scrolled down the file to see that but it is what it is.

PR #669, opened 2023-04-09
Fix Stale References

Roberto-Nessy:
The pointers used to update the target pointer and reticle colours when Navi is set to rainbow were not reset between areas causing them to become stale and edit unintended memory locations which could cause crashes, game freezes, and texture corruptions.


Roberto-Nessy:
Because the adult and child trade quests both used the same variable trading the last adult item wouldn't unequip it so added a bitmask here

PR #673, opened 2023-05-04
Fix Logic with OGC and mixed entrance pools

leggettc18:
Adds a `GANONS_CASTLE_LEDGE` area and a `BuiltRainbowBridge` logic var. Added an event to `GANONS_CASTLE_GROUNDS` that sets `BuiltRainbowBridge` to true. Adds Entrances for `GANONS_CASTLE_LEDGE` to `GANONS_CASTLE_GROUNDS` when `IsAdult` and `BuiltRainbowBridge` (or the glitched logic that lets you skip the rainbow bridge) are true, `HYRULE_CASTLE_GROUNDS` when `IsChild` is true, and `GANONS_CASTLE_ENTRYWAY` When IsAdult is true (if child enters `GANONS_CASTLE_LEDGE` in game they are in `HYRULE_CASTLE_GROUNDS` and thus cannot go back into GANONS_CASTLE_ENTRYWAY). Adds Entrance for `GANONS_CASTLE_GROUNDS` to `GANONS_CASTLE_LEDGE` when `BuiltRainbowBridge` is true, or when the glitched logic that lets you skip the Rainbow Bridge is true.

EDIT: Added a `CASTLE_GROUNDS_FROM_GANONS_CASTLE` to handle age-checking and sending Child Link to `HYRULE_CASTLE_GROUNDS` and Adult Link to `GANONS_CASTLE_LEDGE`. `GANONS_CASTLE_LEDGE` then does not have to worry about Link's Age for it's entrances to `GANONS_CASTLE_GROUNDS` or going back to `GANONS_CASTLE_ENTRYWAY`


leggettc18:
Confirmed, if I hardcode SoT on `OGC_GS` and turn on mixed entrance pools and Closed Door of Time, seeds fail to generate. Without hardcoding Song of TIme, seeds generate just fine, indicating that this resolves the issue.

leggettc18:
Also confirmed if I hardcode SoT on `OGC_GS` on the main branch, the seed _does_ generate successfully, despite the fact that seed should be unbeatable.

source/location_access/locacc_ganons_castle.cpp
```diff
@@ -17,7 +17,7 @@ void AreaTable_Init_GanonsCastle() {
                  // Exits
                  Entrance(GANONS_CASTLE_LOBBY, { [] { return Dungeon::GanonsCastle.IsVanilla(); } }),
                  Entrance(GANONS_CASTLE_MQ_LOBBY, { [] { return Dungeon::GanonsCastle.IsMQ(); } }),
-                 Entrance(GANONS_CASTLE_GROUNDS, { [] { return true; } }),
+                 Entrance(CASTLE_GROUNDS_FROM_GANONS_CASTLE, { [] { return true; } }),
```
Kewlan, line 20:
```suggestion
                 Entrance(HYRULE_CASTLE_GROUNDS, { [] { return IsChild; }}),
                 Entrance(GANONS_CASTLE_LEDGE, { [] { return IsAdult; }}),
```
Could this work instead? If so we don't need the extra `CASTLE_GROUNDS_FROM_GANONS_CASTLE` area.

leggettc18: I had that at one point, and it would work just fine, but from a style perspective personally I like having the intermediate area to handle the age-gating, similar to how the `CASTLE_GROUNDS` area works currently. Although I suppose the reason `CASTLE_GROUNDS` exists is for re-use, whereas `CASTLE_GROUNDS_FROM_GANONS_CASTLE` is only used once (at least for now).

Archez: It is also useful as the entrance pairs listed in `entrance.cpp` can only point to one region that applies to both ages. So the age gating helps with entrance shuffle too. Otherwise `GANONS_CASTLE_LEDGE` would have to have age checks for every access.
I also think it is more readable as child link is never truly able to be at "ganons castle ledge"

leggettc18: Are we OK keeping it as is or do you want me to change it? No rush, just clearing up whether or not this is waiting on me.

Kewlan: I'm fine with keeping it as is, so it's on the maintainers now.
I suspect you'll need to run clang format though. We have a workflow you can do it from if you can't yourself.

leggettc18: Oh I should have it installed, I'll check on that.

PR #681, opened 2023-06-02
Dampe progressive hookshot hint

VeryTwistedSpoon:
- Dampe's Diary will now reveal the region of one of the two available progressive hookshots. 
- Added Dampe Hint toggle to Misc Hints list and ensured functionality of the Misc options `ToggleSet` so all Misc hints can still be toggled on or off at once.

Dampe Diary

- [x] Hint text needs localisation for French, Spanish, Italian and German


HylianFreddy:
This is cool! I have just a couple of suggestions.

The hookshotLocation vector in CreateDampeHint() can be empty if the player starts with hookshot, so you have to add a check for it to avoid crashes.
You can copy how it's done for light arrows in CreateGanonText().

Regarding the dialog, I think it would be better to keep the original text and only change the location name, like in the OoT Randomizer:
"Whoever reads this, please enter <location>. I will let you have my stretching, shrinking keepsake.^I'm waiting for you.&--Dampé"
To put the location name in the middle you can add another hint to the HintTable (see how the merchants hints do this).

For the other languages, you can ask in the Discord server for translations. Please don't use machine translations, as they're often wrong.

VeryTwistedSpoon:
Implemented into https://github.com/gamestabled/OoT3D_Randomizer/pull/690 with enhancements.

PR #690, opened 2023-07-23
House of Skulltula and Dampé's diary hints

HylianFreddy:
The new skulltula hints will make the cursed men say what reward they'll give once their curse is lifted.
Currently all 5 hints are toggled with the same setting, but in the OoT Randomizer they have 5 different settings instead. If that is preferred I can add those.

skulltula_hint

I've also added an overload for the Text::Replace function, so the string to insert can be a Text object instead.

This PR is built on top of the currently open #681, and includes fixes for the problems I mentioned in a comment on that PR.
I'll leave this one as a draft for now.


HylianFreddy:
I'll mark this PR as ready.
If it gets merged as is, then #681 should get closed because its commits are included here.
If instead the other PR is fixed and merged, please do not merge this one until I rebased it.

VeryTwistedSpoon:
Sorry it took so long for me to implement your suggestions. I had done everything except the italian and german. After seeing your code I think its better to leave it part of this pull req. My code ended up devolving into 2 split dialogs to incorporate the original game text but your string replace and Text overload solution is much more appealing.

PR #691, opened 2023-07-23
Make dungeon rewards hints available on gossip stones if they're not on altar or compasses

HylianFreddy:
A simple change to skip setting medallions as already hinted if they don't appear on ToT altar or compass messages, so they can appear on gossip stones instead.


HarmfulBreeze:
Just giving my two cents here; it looks a little odd to me that dungeon rewards would be marked as hinted in the `BuildDungeonRewardText` function in the first place. Shouldn't they be marked as hinted somewhere else, like in `CreateCompassAndGearMenuHints` and `CreateAltarText`?

gymnast86:
I think setting them as hinted here is fine since this is the function where the hint text is created.

PR #709, opened 2023-11-27
Revert "Destroy the hintstones on Citra as they cause huge lag"

raphaelthegreat:
The slowdown has been fixed on upstream


Kewlan (approved):
This gave me a good chuckle. Funny to see a big time Citra dev here. Thanks!

PR #718, opened 2024-03-03
Add Enemy Soul Shuffle

HylianFreddy:
This feature is based on the homonymous one from OoTR's Dev-Rob branch, but it's a bit different. When enabled, enemies will become invincible until their soul is obtained (instead of not spawning at all). A new "Enemy Soul" item will be added to the pool for each enemy type. For example, all Tektites will become vulnerable when the Tektite Soul is obtained.

#### Appearance
Soulless enemies will appear as purple flames. Initially I wanted to apply some sort of graphic filter to their models, but I don't think I'll be able to do that without more decomp knowledge.

The Soul item model is a recolored version of the shop fairy item. I chose this model because it doesn't normally appear outside shops, so it should be easy to recognize.

#### Logic
I have added souls requirements to the location access files. I took a small shortcut for Business Scrubs and Gold Skulltula locations by adding the requirement to the `GetConditionsMet` function, where they can be recognized through their Category.

One thing missing from this PR is Logic for Master Quest dungeons. I have tried adding it but I quickly lost motivation because I don't know the MQ dungeons by memory, and I can't just copy OoTR logic because of the many new areas that have been added there. Maybe a future PR could port all those new areas and make it easier to add enemy souls logic.
I have added a check in the `ValidateSettings` function to prevent seed generation if MQ Dungeon Count is not zero (and I also modified the function to make it properly account for randomized settings).

#### Menus
I made a submenu in the starting inventory for the Enemy Souls. To make the menu unselectable when it's not relevant, I added the "Lock" feature to the Menu class, copied from the Option class.

The in-game menu will have an extra page listing all the souls and marking those that have been collected. I have copied the style of this page from @Kewlan's custom branch, I hope that's OK.

#### Technical notes
Soulless enemies are made invulnerable by disabling any collision with their AC colliders, making any AT colliders that touch them simply go through them without interacting. This means that even the hookshot won't connect when aimed at gibdos/freezards/like-likes/etc.
Collision with the enemy's AT and OC colliders is unaffected, so their attacks can still be blocked with a shield.
For some enemies that can normally die without being attacked directly I have added more specific patches (e.g. Anubis in a fire circle).

#### Potential changes / things to do
- I have chosen to combine some enemies into the same soul (like Flying Pot and Flying Floor Tile into the Flying Trap Soul), but they can be split fairly easily if desired.

#### Preview video
https://github.com/gamestabled/OoT3D_Randomizer/assets/82058772/aba51436-effd-4e4b-b8cf-339d2da5c13a

#### Preview 3DSX builds
~~Build for this branch (commit 0162158a):~~
[~~OoT3D_Randomizer_0162158a.zip~~](https://github.com/gamestabled/OoT3D_Randomizer/files/14472380/OoT3D_Randomizer_0162158a.zip)

~~Build including all my other features, as visible from commit 43063df2:~~
[~~OoT3D_Randomizer_43063df2.zip~~](https://github.com/gamestabled/OoT3D_Randomizer/files/14472383/OoT3D_Randomizer_43063df2.zip)

Build for this branch (commit 72819d44):
[OoT3D_Randomizer_72819d44.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/14608539/OoT3D_Randomizer_72819d44.zip)

Build including all my other features, as visible from commit 39e4665b:
[OoT3D_Randomizer_39e4665b.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/14608541/OoT3D_Randomizer_39e4665b.zip)


Kewlan:
> I hope that's OK

It is very ok! Good choice. 👍

PR #720, opened 2024-03-03
Add Triforce Hunt

HylianFreddy:
This feature is the same as the homonymous one from OoTR. When enabled, a configurable number of Triforce pieces will be added to the pool, and the game will end when the required amount is obtained.

Ganon's Boss Key is removed from the game, and the Ganon location (considered as the "win condition") is logically accessible from the world root when all the Triforce pieces have been collected. This means that, like in OoTR, the logic will always expect all pieces to be required, regardless of the "Required Pieces" setting.

When the Triforce is complete, the game will be saved and the ending sequence will play. The next time the file is opened, Ganon's Boss Key will be given immediately.

#### Settings changes
Because the piece count settings have options from 1 to 200, I have added a "fast scrolling" triggered by holding A while pressing left and right. To do this I also had to merge together the functions to scroll the options into one `ScrollOptionIndex`, I hope that's not bad style.

For the Ganon BK setting, I have added an unselectable "Triforce" option that is only showed when Triforce Hunt is enabled and the setting is locked. I know this is a bit hacky but I couldn't find a better way to show the setting as locked and avoid checking for triforce hunt every time the GanonBK setting is used for anything.

#### Text
I have added a system to add custom control codes to the game text. I made it to show how many Triforce pieces the player has when they find a new one. This will also allow to create more dynamic text in the future if it's ever needed.
I have also removed some of the existing quick text patches and reimplemented them in the new C functions.

#### Potential changes / things to do
- ~~When playing in multiplayer, I think only the player that gets the final piece will warp to the credits. I'm not sure if this should be changed.~~ This has been fixed so that all players will warp to the credits.

#### Preview video
https://github.com/gamestabled/OoT3D_Randomizer/assets/82058772/0d7dcebd-dbee-4789-8230-6d5c327b406a

#### Preview 3DSX builds
~~Build for this branch (commit 721be0eb):~~
[~~OoT3D_Randomizer_721be0eb.zip~~](https://github.com/gamestabled/OoT3D_Randomizer/files/14472421/OoT3D_Randomizer_721be0eb.zip)

~~Build including all my other features, as visible from commit 43063df2:~~
[~~OoT3D_Randomizer_43063df2.zip~~](https://github.com/gamestabled/OoT3D_Randomizer/files/14472383/OoT3D_Randomizer_43063df2.zip)

Build for this branch (commit 96bd9e06):
[OoT3D_Randomizer_96bd9e06.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/14608576/OoT3D_Randomizer_96bd9e06.zip)

Build including all my other features, as visible from commit 39e4665b:
[OoT3D_Randomizer_39e4665b.zip](https://github.com/gamestabled/OoT3D_Randomizer/files/14608541/OoT3D_Randomizer_39e4665b.zip)


Kewlan:
For multiplayer, maybe the behaviour should change to something like this:
- Whether it's singleplayer or multiplayer, it checks for when you have gone from having less than required to required triforce pieces or higher (to account for two get one at once), then wait to teleport to credits.
- The teleport happens as soon as the player can move freely again. This is so it doesn't mess anything up for other players that are in the middle of something important, for example, after setting a chest opened bit but before the item has actually been acquired.

HylianFreddy:
I have implemented the credits warp for multiplayer. For the player that gets the final piece, the warp will start when the text ends. If instead the final triforce pieces are received from the network, the warp will start when the game/player isn't busy.

Kewlan:
Can you add `TRIFORCE_PIECE` to ignored playthrough hint items here, so gossip stones don't get filled with triforce pieces:
https://github.com/gamestabled/OoT3D_Randomizer/blob/3d5291ffcea9562e9243a951fde6a0a49f104143/source/hints.cpp#L902-L909
I think they can always be ignored because you can still tell based on the given hints where you should and don't need to search for pieces.

HylianFreddy:
> Can you add `TRIFORCE_PIECE` to ignored playthrough hint items here, so gossip stones don't get filled with triforce pieces:
> 
> https://github.com/gamestabled/OoT3D_Randomizer/blob/3d5291ffcea9562e9243a951fde6a0a49f104143/source/hints.cpp#L902-L909
> 
> 
> I think they can always be ignored because you can still tell based on the given hints where you should and don't need to search for pieces.

Done!

gymnast86 (approved):

PR #725, opened 2024-04-05
Multiplayer: Show up to 3 other players as Link

Kewlan:
https://github.com/gamestabled/OoT3D_Randomizer/assets/5352197/379219c8-57fd-4f3d-bb2b-70eeaf115159

- If too many extra Links are drawn, the game gets graphical glitches that last until the game is restarted, so limit the amount to three. The extra Links will be shown as flames like before.
- On 64, Link's joint table is 22 in length of Vec3s. Here on 3DS I don't actually know but just assumed that it's still Vec3s. However, to get the full animation here we need to copy 217! Maybe there's a better way to set the pose with an animation index and current frame? Regardless as of now just for the posing it'll send 1300 bytes per frame per player. Most likely this will have to be skipped for when playing on console, so Links will just be T-posing.
- Puppet Child Link works weird for some reason. The position is off when standing and climbing. ~I don't know how to fix.~ Workaround by moving the actor's position down 12 units. There were other issues that happened along side this before, but they don't seem to happen any longer?
- ~Custom tunic colors mess up the graphics on the puppet's tunic.~

### DIFF
--- a/code/src/actors/link_puppet.c
+++ b/code/src/actors/link_puppet.c
@@ -11,16 +11,36 @@
 #define MAX_PUPPETS_AT_ONCE 3
 
 ActorInit EnLinkPuppet_InitVars = {
-    0x1,                  // ID
-    ACTORTYPE_NPC,        // Type
-    0xFF,                 // Room
-    0b110000,             // Flags
-    21,                   // Object ID (20: Adult, 21: Child)
-    sizeof(EnLinkPuppet), //
-    EnLinkPuppet_Init,    //
-    EnLinkPuppet_Destroy, //
-    EnLinkPuppet_Update,  //
-    EnLinkPuppet_Draw,    //
+    0x1,                             // ID
+    ACTORTYPE_NPC,                   // Type
+    0xFF,                            // Room
+    0x2000410,                       // Flags
+    21,                              // Object ID (20: Adult, 21: Child)
+    sizeof(EnLinkPuppet),            //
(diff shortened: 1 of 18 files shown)


code/src/actors/link_puppet.c
```diff
@@ -0,0 +1,147 @@
...
+                                        void* param_5, u32 animation, s32 limbBufCount, void* jointTable,
+                                        void* morphTable);
+#define SkelAnime_InitLink ((SkelAnime_InitLink_proc)0x3413EC)
+
+void EnLinkPuppet_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnLinkPuppet* this = (EnLinkPuppet*)thisx;
+
+    this->base.room = -1;
+
+    SkelAnime_InitLink(&this->skelAnime, PLAYER->zarInfo, globalCtx, PLAYER->cmbMan, thisx->unk_178, 0, 9, NULL, NULL);
```
HylianFreddy: I think the custom tunic glitches would be fixed if, after this call to `SkelAnime_InitLink`, you do something similar to what is done in `Player_SetChildCustomTunicCMAB`.
The only things to change should be the Texanim value (to match Link's current age) and replacing `PLAYER->skelAnime` with `this->skelAnime`.

Kewlan: Thanks!

gymnast86 (approved):

PR #727, opened 2024-04-27
Improve EUR address handling

Kewlan:
- ### Combine linker scripts
Pass `_EUR_` to linker script when building EUR patch and use built-in function `DEFINED` to determine the address.
- ### Use macro for game addresses in C code
Takes the USA address and when building for EUR, adds the necessary offset.
I decided to wrap all references to game addresses in it, and also remove the `*_addr` defines for consistency.


code/include/z3D/z3D.h
```diff
@@ -10,6 +10,21 @@
...
+    #define GAME_ADDR(addr)                                    \
+        ({                                                     \
+            int finalAddr = (addr);                            \
+            if (addr >= 0x41A144 && addr <= 0x43668B) {        \
+                finalAddr += 0x24;                             \
+            } else if (addr >= 0x436690 && addr <= 0x4A5ADF) { \
+                finalAddr += 0x20;                             \
+            }                                                  \
+            finalAddr;                                         \
+        })
```
HylianFreddy:
```suggestion
    #define GAME_ADDR(addr)                                       \
        ((addr >= 0x41A144 && addr <= 0x43668B)   ? (addr + 0x24) \
         : (addr >= 0x436690 && addr <= 0x4A5ADF) ? (addr + 0x20) \
                                                  : addr)
```
I believe a macro like this should be computed at compile time, rather than adding a code block that will run each time a base game function is called.
You can achieve this by writing it with ternary operators, which make the compiler realize what you're writing is actually a constant value.

code/Makefile
```diff
@@ -55,21 +55,21 @@ CFLAGS	+=	$(INCLUDE) -DARM11 -D_3DS $(VERFLAGS)
 
 CXXFLAGS	:= $(CFLAGS) -fno-rtti -fno-exceptions -std=gnu++11
 
+LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map) -T $(TOPDIR)/oot.ld $(VERFLAGS) -lgcc
```
HylianFreddy:
```suggestion

LINK_SCRIPT	:= oot.ld

LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map) -T $(TOPDIR)/$(LINK_SCRIPT) $(VERFLAGS) -lgcc
```
You shouldn't delete the `LINK_SCRIPT` variable because the `.elf` target depends on it.
This is needed to make `make` detect changes on the linker script even if it's the only modified file.

code/oot.ld
```diff
@@ -1240,347 +1240,347 @@ SECTIONS
 		*(.patch_SleepQueryCallback)
 	}
 
-	.patch_OcarinaNoteButtonsPress 0x41AAAC : {
+	.patch_OcarinaNoteButtonsPress ((DEFINED(_EUR_)) ? 0x41AAD0 : 0x41AAAC) : {
```
HylianFreddy:
```suggestion
	/* Start of regional differences: 0x41A144 */
	region_offset = DEFINED(_EUR_) ? 0x24 : 0;

	.patch_OcarinaNoteButtonsPress 0x41AAAC + region_offset : {
```
This is only a suggestion, but because your other changes make it so the developer doesn't have to worry about address conversion anymore, maybe it would be good to do the same in the linker script too.
If you add a variable like `region_offset` and set its value at specific points in the file, every patch can just be written as `<address> + region_offset`.

The other two points where the offset value would change are before `patch_GearMenuEmptySlot`:
```
	/* Variation of regional differences: 0x436690 */
	region_offset = DEFINED(_EUR_) ? 0x20 : 0;

	.patch_GearMenuEmptySlot 0x438C00 + region_offset : {
```
and before `patch_OverrideGiDrawIdPlusOne`:
```
	/* End of regional differences: 0x4A5AE0 */
	region_offset = 0;

	.patch_OverrideGiDrawIdPlusOne 0x4BC610 : {
```

HylianFreddy (changes requested):
Good job on this, I just have a couple of suggestions/changes which I put in the comments (and also on this branch if you prefer: https://github.com/HylianFreddy/OoT3D_Randomizer/commits/region-addr)

I also have a question, should all base game addresses be wrapped within the macro? Some still aren't, but it doesn't really change anything so it's fine either way.
I found 43 with this regex: `\)0x[a-f,A-F,0-9]{6}`

Kewlan:
@HylianFreddy Excellent suggestions! And thanks for the commits!

code/include/z3D/z3D.h
```diff
@@ -700,22 +694,22 @@ typedef struct MainClass {
 extern GlobalContext* gGlobalContext;
 extern const u32 ItemSlots[];
 extern const char DungeonNames[][25];
-#define gSaveContext (*(SaveContext*)0x00587958)
+#define gSaveContext (*(SaveContext*)GAME_ADDR(0x587958))
 #define gStaticContext (*(StaticContext*)0x08080010)
```
HylianFreddy, line 698: You missed one address here :)
After that, lgtm!

Kewlan: I intentionally excluded that since it's outside the range that can be seen in Ghidra, so I'm just assuming that plays by different rules than these normal addresses. Same with the other 8-symbol long addresses. Though, those are all above `0x10000000`. Do you think `gStaticContext` should be included anyway?

HylianFreddy: Oh, you're right, I didn't think about that. Let's leave it like this then.

HylianFreddy (approved):

PR #731, opened 2024-05-13
Some text fixes

HylianFreddy:
- Reduce the maximum pixel width for text boxes that contain an item icon. Currently some messages without manual newlines look like this:
text overflow

- Fix Dark Link's name so it consistently uses the player name for all languages in the game's text boxes, while retaining the name "Link" for the spoiler log and custom menus. Currently the `Dark Link Soul` appears as `Dark @ Soul` in shops for non-English languages.


source/custom_messages.cpp
```diff
@@ -1145,7 +1146,14 @@ static std::map<std::string, int> pixelWidthTable = {
...
 // next line in the string from the last calculated newline.
 static size_t NextLineLength(std::string* textStr, size_t lastNewline) {
-    constexpr size_t maxLinePixelWidth = 287;
+    size_t maxLinePixelWidth = 287;
+    size_t lastCaret         = textStr->rfind('^', lastNewline);
+    size_t lastItemIcon      = textStr->rfind("\x7F\x0F", lastNewline);
+
+    // For text boxes with an item icon, decrease max line pixel width to leave space for the icon
+    if (lastItemIcon != std::string::npos && (lastCaret == std::string::npos || lastItemIcon > lastCaret)) {
+        maxLinePixelWidth = 252;
```
Kewlan, line 1155: Normally the text starts 56 pixels from the left of the screen, but with an icon it starts 104 pixels from the left instead, so 48 pixels further in. Any reason you reduced it by 35 pixels instead?

HylianFreddy: I tried some different values and converged on one that fixed all the overflowing texts while changing the least amount of already correct texts.
Reducing the length more would add more newlines where they're not needed, causing more conflicts with the manually inserted ones.

Kewlan (approved):

PR #735, opened 2024-05-15
Correctly handle ExtData when copying savefiles

HylianFreddy:
Currently if you copy a save file to a different slot the extdata are not properly handled, and whatever stale data is in the `gExtSaveData` struct will be saved onto the target slot.
This fixes it so the extdata are properly loaded from the source slot before they get saved on the target slot.

### DIFF
--- a/code/oot.ld
+++ b/code/oot.ld
@@ -744,6 +744,10 @@ SECTIONS
 		*(.patch_ChildCanOpenBowSubMenu)
 	}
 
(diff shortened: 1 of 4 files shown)


UraYukimitsu (approved):

UraYukimitsu:
I think I completely forgot about copying save files when I added extdata support

Kewlan (approved):

PR #737, opened 2024-05-23
Add "Bosses only" option to Enemy Soul Shuffle

HylianFreddy:
This new option allows to play with soulless bosses while keeping all other enemies normal, as was requested.
I've also changed the text for the Gerudo Soul to clarify that it affects Iron Knuckles too. An option to split some souls could be added in the future but it'll require more work.


ScArWlvrne (approved):
I trust Freddy and I want this for Asyncs lol

Roberto-Nessy (approved):

PR #751, opened 2024-08-11
Fix permanent items not being placed

### LINKED ISSUE #745: Second Deku Nut Upgrade was never found
Played with Enimes souls and Ocarina buttons randomized and found everything (including both Deku Stick upgrades) but never got the second Deku Nut one. Please fix this bug! There's only about 5 checks left and none of them in my spoiler log say "Deku Nut Capacity".

Roberto-Nessy:
Changed the 2-way distinction of "is advancement" and "is not advancement" to a 3 way distinction of "is advancement", "is junk", and "neither" so that items such as maps and compasses can be prioritised over junk items in the fast fill section.

Erred on the side of not junk for most items except rupees and ammo refills.


gymnast86 (dismissed):

HylianFreddy:
Hi, I have a suggestion I made on Discord some time ago but then we never talked about it.
I think the function `PlaceJunkInExcludedLocation` should be changed to prioritize placing Junk items over non-advancement items, to avoid placing things like stick/nut upgrades or maps/compasses in excluded locations.

For example  by repeating the loop to look for true junk before non-advancement items, like this:
```
void PlaceJunkInExcludedLocation(const LocationKey il) {
    // place a junk item in this location
    for (size_t i = 0; i < ItemPool.size(); i++) {
        if (ItemTable(ItemPool[i]).IsJunk()) {
            PlaceItemInLocation(il, ItemPool[i]);
            ItemPool.erase(ItemPool.begin() + i);
            return;
        }
    }
    // if no junk item was found, try placing a non-advancement item
    for (size_t i = 0; i < ItemPool.size(); i++) {
        if (!ItemTable(ItemPool[i]).IsAdvancement()) {
            PlaceItemInLocation(il, ItemPool[i]);
            ItemPool.erase(ItemPool.begin() + i);
            return;
        }
    }
    printf("ERROR: No Junk to Place!!!\n");
}
```

HylianFreddy (approved):
I have pushed a commit with my suggestion above with permission from Nessy.

PR #766, opened 2025-03-25
Enemy Randomizer

HylianFreddy:
This PR adds a new setting that randomizes most enemies in the game, like in the other OoT randomizers.

The first commit includes various updates to the object handling system that were done in preparation for the new feature:
- The base game code will automatically look for objects in the Extended Context if it doesn't find them in the normal one.
- All objects required by randomizer features will now be spawned in the Extended Context, and only if they're not already in the normal one.
- On room change, all non-persistent objects will be cleared from the Extended Context to prevent it from running out of space when adding objects for randomized actors.
- On scene load, objects can be spawned "persistently" so they won't unload on room change. Currently this is done for `zelda_gi_melody` (custom assets) and `zelda_dangeon_keep` (dungeon objects, ice trap chests, etc.).
- Ice Trap chests will spawn the Freezard object for the ice smoke effect, which was missing until now.

While updating the object code for Skulltula Randomizer, I removed the `GsQueue` as it was only being filled on room loads anyway. Now all random GS will be spawned after the Actor List command. In my test playthrough I didn't notice any problems with this, but maybe @Kewlan knows of some other reason for the queue to stay.

The second commit adds the actual Enemy Randomizer code. The feature is usable but not yet complete, so I'm opening the PR as draft to allow people to give suggestions and play-test the attached build if they want to.

Currently I plan to:
- [x] Add logic to guarantee having the items required to clear rooms.

Other potential things that could be done:
- Patch and add more enemies to the randomization options (e.g. Poe Sisters, Gerudo Fighters,...)
- Add more settings to customize the experience further (e.g. exclude some enemy types, or mini-boss locations,...)

~~3DSX build:
[OoT3D_Randomizer_1a4d4f58.zip](https://github.com/user-attachments/files/19455819/OoT3D_Randomizer_1a4d4f58.zip)~~

A playable build can be downloaded from my fork's [release page](https://github.com/HylianFreddy/OoT3D_Randomizer/releases/latest).


Kewlan:
As I recall, the GS queue was just my way of handling the spawning of them, so if they still always appear then I approve of the change. 👍

Kewlan (approved):

PR #792, opened 2026-01-09
Change ExtInf to a struct

HylianFreddy:
This changes the type of `extInf` from a u8 array to a generic struct, in order to make it easier to add bitfields containing more than 8 flags, without having to add more functions to the multiplayer implementation because anything added to `extInf` will be synced automatically.

I also moved the `hasTimeTraveled` flag out of the struct because it was intentionally skipped in the syncing code.


Kewlan (approved):
Excellent use case for a union!

Issue #571, opened 2022-10-14
[Feature Request] Black Rupees that make you lose rupees like in the other games


LinkexaYT:
If possible, I think there should be Rupoors like in Four Swords, Phantom Hourglass and Skyward Sword, and everytime you collect one, you lose rupees. The rupoors should be found in small chests, since I don't think it's possible to make it part of a drop or as a freestanding one like the rupees in Kokiri Forest.

PR #799, opened 2026-03-01
Add Rupoor Trap

BuffYoda21:
This adds a new toggleable trap that replicates the effects of a Rupoor from other zelda titles. In addition to being toggleable, it also has 3 "severities" you can set it to:
- 10, which acts like a vanilla rupoor from other zelda titles deducting 10 rupees from your wallet.
- Random Ratio, which deducts a random percentage (between 5-65%) of your current *max* rupees.
- Bankruptcy, which takes all of your rupees and makes you sad.


BuffYoda21:
Also may close #571 though there is no custom model associated since it is designed to function like all the other traps.

HylianFreddy:
Good job! Would you also be interested in adding a model for this if I helped with it, or are you satisfied for now?

BuffYoda21:
That'd be great! Was initially planning on doing that but later realized that because I was designing it like all the other custom ice traps it would be easier to just not do that and it would still look fine. That said, having link hold up a Rupoor during the "FOOL!" text would immediately give away what trap you actually got. If we were to add a model maybe adding a special case to change the FOOL dialogue to something else would make it feel like an entirely separate item? Or maybe even reworking it as an entirely new item? Or am I overthinking this?

HylianFreddy:
Yeah I was thinking of making a new item for the Rupoor, so the text could be changed. Though integrating it with the other trap types will require reworking the existing system a bit. I'll see what we can do for that.

BuffYoda21:
It does seem like having it be it's own item would be the cleanest solution. Obviously still keeping it toggleable though

BuffYoda21:
If we do go add an entirely new item, would it make sense to close this PR without merging and open a new one since just about everything changed in this one would no longer apply?

HylianFreddy:
Well no, the core feature is already in this PR (the trap effect and the app setting).

What I'm thinking is to make a new GI item to take care of the model, but instead of adding it to the app side as a shuffleable item we can change the IceTrap system to detect when the random trap would be a Rupoor and give the new item instead of the trap item. I haven't implemented this yet though.

If you want, I can merge this PR and then later make a new one for the model. Or we can keep this one open and I'll make a PR to your fork. Tell me what you prefer.

BuffYoda21:
Oh ok I get what you mean now. Yea I'd think that this right now is functionally complete and there wouldn't be any harm to merging now and we can add the model later.

HylianFreddy:
Alright, I will merge this PR then, thank you!

HylianFreddy (approved):

PR #800, opened 2026-03-04
Add item and model for Rupoor trap

### LINKED ISSUE #571: [Feature Request] Black Rupees that make you lose rupees like in the other games

HylianFreddy:
This adds a black rupee model for the Rupoor trap, as discussed in PR #799.
I made some updates to the IceTrap system to allow giving custom items as traps, and implemented the Rupoor as a standalone item that will only be given by ice traps whose random effect is the rupoor.

https://github.com/user-attachments/assets/b641bcbc-683e-49df-bff1-515943e71d84


code/src/icetrap.c
```diff
@@ -96,22 +96,33 @@ void LinkDamageNoKnockback(void) {
...
+                       : possibleItemTraps[hash % possibleItemTrapsAmount];
+}
 
-        u8 trapType = ICETRAP_VANILLA; // Default to ice trap
-        if (gSettingsContext.randomTrapDmg != RANDOMTRAPS_OFF) {
-            trapType = possibleItemTraps[pRandInt % possibleItemTrapsAmount];
+void IceTrap_UpdateOverride(ItemOverride* override, u8 isFromChest) {
+    if (override->value.itemId == GI_ICE_TRAP) {
+        IceTrap_ActiveHash = Hash(override->key.all);
+        if (IceTrap_GetType(IceTrap_ActiveHash, isFromChest) == ICETRAP_RUPOOR) {
```
VictorRemmerswaalAuxilium, line 111: Maybe, for future proofing make this a function like "IsIceTrapWithItem()"?

HylianFreddy: ~That function doesn't return a boolean, so naming it "Is..." would be confusing. Why do you think the current name is bad? Sure it's pretty generic, but I'm not understanding what you mean by "future proofing"...~

HylianFreddy: Oh, I get it now. Sorry I thought you were talking about `IceTrap_UpdateOverride` because of how GitHub displayed your comment.

I still don't think adding a `IsIceTrapWithItem` would improve the code, there must still be an equality check for the ICETRAP_RUPOOR effect specifically, and if more of these effects are added in the future there will be a `switch` or some `else if`s, so the extra check would be wasted.

Roberto-Nessy (approved):

PR #801, opened 2026-03-06
Add Personalization Presets

### LINKED ISSUE #759: Add Ability to save Cosmetic Settings directly to file from the Randomizer Configuration Settings
I have a few different cosmetic options that I like to switch between but it is a little tedious to manually adjust them all from settings or to copy and paste the file every time. Adding something similar to the custom presets feature but for cosmetic settings would be nice.

BuffYoda21:
This reworks existing preset saving logic to also work for personal/cosmetic settings to allow easier switching between them. #759

### DIFF
--- a/source/menu.cpp
+++ b/source/menu.cpp
@@ -24,6 +24,7 @@ bool chosePlayOption;
 u16 pastSeedLength;
 PrintConsole topScreen, bottomScreen;
 std::vector<std::string> presetEntries;
+std::vector<std::string> cosmeticPresetEntries;
 std::vector<Menu*> menuList;
 Option* currentSetting;
 Menu* currentMenu;
@@ -150,6 +151,9 @@ void MoveCursor(u32 kDown, bool updatedByHeld) {
         } else if (currentMenu->mode == LOAD_CUSTOM_PRESET ||
                    currentMenu->mode == DELETE_CUSTOM_PRESET) { // Number of presets if applicable
             max = presetEntries.size();
+        } else if (currentMenu->mode == LOAD_COSMETIC_PRESET ||
+                   currentMenu->mode == DELETE_COSMETIC_PRESET) { // Number of presets if applicable
+            max = cosmeticPresetEntries.size();
         } else if (currentMenu->mode == GENERATE_MODE) { // Generate menu: 2 options
             max = 2;
         } else if (currentMenu->itemsList != nullptr) {
@@ -183,7 +187,8 @@ void MoveCursor(u32 kDown, bool updatedByHeld) {
 
         // Scroll Check
         u16 max_entries_on_screen = MAX_SUBMENUS_ON_SCREEN;
-        if (currentMenu->mode == LOAD_CUSTOM_PRESET || currentMenu->mode == DELETE_CUSTOM_PRESET) {
+        if (currentMenu->mode == LOAD_CUSTOM_PRESET || currentMenu->mode == DELETE_CUSTOM_PRESET ||
+            currentMenu->mode == LOAD_COSMETIC_PRESET || currentMenu->mode == DELETE_COSMETIC_PRESET) {
             max_entries_on_screen = MAX_SUBMENU_SETTINGS_ON_SCREEN;
         }
         if (currentMenu->menuIdx >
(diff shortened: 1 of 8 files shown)


BuffYoda21:
If I missed anything please let me know

BuffYoda21:
Naming scheme probably also needs to be adjusted to avoid confusion but I'll wait on approval for that.

source/menu.cpp
```diff
@@ -361,13 +380,18 @@ void UpdatePremadePresetsMenu(u32 kDown) {
...
+    const u8 DELETE                   = presetType == SETTINGS ? DELETE_CUSTOM_PRESET : DELETE_COSMETIC_PRESET;
+    std::vector<std::string>& entries = presetType == SETTINGS ? presetEntries : cosmeticPresetEntries;
+    OptionCategory category           = presetType == SETTINGS ? OptionCategory::Setting : OptionCategory::Cosmetic;
+
     consoleSelect(&topScreen);
     // clear any potential message
     ClearDescription();
-    if (kDown & KEY_A && currentMenu->mode == LOAD_CUSTOM_PRESET && !presetEntries.empty()) {
+    if (kDown & KEY_A && currentMenu->mode == LOAD && !entries.empty()) {
         Settings::SetDefaultSettings();
```
HylianFreddy: This function also needs to be updated, otherwise it will reset all normal settings when loading a cosmetic preset. I suggest adding a parameter like this:
```
void SetDefaultSettings(bool forCosmetics = false);
```
and then changing the inner loop to
```
for (auto option : *menu->settingsList) {
    if (forCosmetics == option->IsCategory(OptionCategory::Cosmetic)) {
        option->SetToDefault();
    }
}
```

source/menu.hpp
```diff
@@ -32,20 +35,25 @@
 #define CYAN "\x1b[36m"
 #define WHITE "\x1b[37m"
 
+typedef enum {
+    SETTINGS,
+    COSMETICS
+} PresetType;
+
```
HylianFreddy: I suggest using an enum class here for clarity, so the values will be namespaced like `PresetType::SETTINGS`.
```suggestion
enum class PresetType {
    SETTINGS,
    COSMETICS,
};
```

HylianFreddy (changes requested):

BuffYoda21:
I think we should also change all instances of "CUSTOM" to "SETTINGS" for clarity.
Ex: `LOAD_CUSTOM_PRESET` -> `LOAD_SETTING_PRESET`

HylianFreddy:
"Custom" presets are opposed to the "premade" ones built into the app, so the word "custom" should stay I think, but adding "Settings" would be good.

For even more clarity, if you want you could divide the options into 2 menus to make a structure like this:
- Settings Presets
    - Randomization Settings
        - Load Premade Preset
        - Load Custom Preset
        - Save Custom Preset
        - Delete Custom Preset
        - Reset to Default Settings
    - Cosmetic Settings
        - Load Custom Preset
        - Save Custom Preset
        - Delete Custom Preset
        - Reset to Default Settings

HylianFreddy (dismissed):
This is good for me, but let me know if you want to implement the extra menu.

BuffYoda21:
> "Custom" presets are opposed to the "premade" ones built into the app, so the word "custom" should stay I think, but adding "Settings" would be good.
> 
> For even more clarity, if you want you could divide the options into 2 menus to make a structure like this:
> 
> * Settings Presets
>   
>   * Randomization Settings
>     
>     * Load Premade Preset
>     * Load Custom Preset
>     * Save Custom Preset
>     * Delete Custom Preset
>     * Reset to Default Settings
>   * Cosmetic Settings
>     
>     * Load Custom Preset
>     * Save Custom Preset
>     * Delete Custom Preset
>     * Reset to Default Settings

This could be cool and while I initially was thinking it might be more trouble than it is worth, I am not sure how else to make that clear distinction between both premade/custom and settings/cosmetic at the same time without making the menu look overly bloated. I guess it is probably worth looking into.

BuffYoda21:
> "Custom" presets are opposed to the "premade" ones built into the app, so the word "custom" should stay I think, but adding "Settings" would be good.

When you say this, do you mean just the user facing side or also in the source code? My main concern is I don't like how long `LOAD_CUSTOM_SETTINGS_PRESET` looks. Though now that I think about it, dividing it into two menus probably means that `LOAD_CUSTOM_PRESET` can stay and handle both settings and cosmetics after some reworking.

HylianFreddy:
I was mainly thinking of the UI, but yeah by making a new menu you can keep the same text and the same define in the code

source/menu.cpp
```diff
@@ -361,13 +391,21 @@ void UpdatePremadePresetsMenu(u32 kDown) {
     }
 }
 
-void UpdateCustomPresetsMenu(u32 kDown) {
+void UpdateCustomPresetsMenu(u32 kDown, PresetType presetType) {
+    const u8 LOAD =
+        presetType == PresetType::RANDOMIZATION ? LOAD_CUSTOM_RANDOMIZATION_PRESET : LOAD_CUSTOM_COSMETIC_PRESET;
+    const u8 DELETE =
+        presetType == PresetType::RANDOMIZATION ? DELETE_CUSTOM_RANDOMIZATION_PRESET : LOAD_CUSTOM_COSMETIC_PRESET;
```
HylianFreddy:
```suggestion
        presetType == PresetType::RANDOMIZATION ? DELETE_CUSTOM_RANDOMIZATION_PRESET : DELETE_CUSTOM_COSMETIC_PRESET;
```

HylianFreddy (changes requested):
Aside from the typo I commented on, the rest looks good to me, good job! Will this be ready for merging after that last fix?

BuffYoda21:
Yea this should be all set now

HylianFreddy (approved):

PR #805, opened 2026-03-17
Update Multiplayer Full Sync

HylianFreddy:
This changes the Multiplayer Full Sync to work correctly with any size of the `ExtInf` struct, by splitting it up into parts of at most 0x400 bits to be synced individually. The amount of parts isn't hardcoded, it will be dynamically computed at compile time. The `mp_completeSyncs` array is expanded to include all the ExtInf parts, and the type of the `neededPacketsMask` sent over the network is changed to a struct of the correct size to have at least 1 bit for each FullSync part.

On the current build there is only 1 ExtInf part, but I tested increasing the size to have more parts and it would look like this:

https://github.com/user-attachments/assets/335a67ba-f9b1-4220-8199-7ffaff00b58c

### DIFF
--- a/code/src/common.h
+++ b/code/src/common.h
@@ -10,6 +10,7 @@
 #define ABS(x) ((x) >= 0 ? (x) : -(x))
 #define MIN(x, y) (x < y ? x : y)
 #define MAX(x, y) (x > y ? x : y)
+#define TMP_ZERO_BUF(size) ((char[size]){ 0 })
 
 #define TICKS_PER_SEC 268123480
 #define SEQ_AUDIO_BLANK 0x1000142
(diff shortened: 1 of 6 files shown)


Kewlan (dismissed):
Incredible. If you're also sure there won't be any off-by-one issues between each extInf sync parts, then this is fully approved!

code/src/common.h
```diff
@@ -10,6 +10,7 @@
 #define ABS(x) ((x) >= 0 ? (x) : -(x))
 #define MIN(x, y) (x < y ? x : y)
 #define MAX(x, y) (x > y ? x : y)
+#define TMP_ZERO_BUF(size) ((char[size]){ 0 })
```
viccie211: I think this name could be better 
```suggestion
#define TMP_ZEROED_BUFFER(size) ((char[size]){ 0 })
```

viccie211 (approved):

code/src/common.h
HylianFreddy: Done!

Kewlan (approved):

PR #807, opened 2026-03-29
Add json library and convert description table into json

BuffYoda21:
This PR adds [this](https://github.com/nlohmann/json) json library that supports jsonc (which is +25k of this PR's diff) and goes ahead and converts all the descriptions originally defined in `descriptions.cpp` to json in `description_table.jsonc`. There are also `item_table.jsonc`, `location_table.jsonc`, and `settings_table.jsonc` which are not implemented yet and are really just included in this PR so I don't lose them. 

I chose jsonc over json since these files would become the primary way developers would interact with surface level changes like adding a new item, setting, location, etc so making them more humanly readable is important. I understand that parsing jsonc vs json does take more overhead but the difference in runtime is all just startup cost and most users likely wouldn't notice a difference.

These data files are intended to be shared with the upcoming OoT3D archipelago implementation so that syncing changes between them remains easy. Best case scenario becoming entirely automated when a new release is published here.

Small note, in order for intellesense not to get upset for trying to reference the json files, contributors will need to add `"${workspaceFolder}/build"` to their includePath in `c_cpp_properties.json`. Not sure if there is another better way to do this since that will mean if you haven't built the app yet or run `make clean` intellesense will get upset.

-------------------- Original Post --------------------

Very big change but I believe this is the best way to ensure maintainability and scalability as the project grows. The json files used here can also be used for an eventual archipelago implementation making maintaining parody between the standalone and AP randomizers much easier. Currently plan to convert the following to be data driven (but will add to this list as necessary):

* Item table (Json written, not implemented)
* Location table (Json written, not implemented)
* Settings (Json written, not implemented)
* Location access (not written)
* Entrance table (not written)
* Descriptions (Json written and implemented)


BuffYoda21:
Went ahead and pushed commit 3b5823d and a32fa39 for the sake of having them here. It currently has an issue that causes a crash during seed generation a few frames after displaying `Caching Settings...Done` that I have been struggling to resolve. Azahar logs print repeated errors similar to `[   4.967245] HW.Memory <Error> core/memory.cpp:Read:481: unmapped Read32 @ 0x00000000 at PC 0x001FC5C8` and when running `$DEVKITARM/bin/arm-none-eabi-addr2line -e OoT3D_Randomizer.elf 0x001FC5C8` it outputs `C:/devkitPro/devkitARM/arm-none-eabi/include/c++/15.1.0/bits/stl_vector.h:1283` but I have no idea how to figure out what vector is the issue other than blindly guessing.

Just to clarify, everything works perfectly up until commit a3fe894

BuffYoda21:
Decided to pivot my current approach and reset basically everything to how it was before I messed with it but keep the json files I made. This is because the way I was doing things (using a python script at the start of the build process to convert the json data into cpp code) was a little dumb and wouldn't have been great to maintain anyway. My new approach will have the json files embedded into the app it's self and read from them on app startup.

BuffYoda21:
Might as well take this opportunity to update my fork since there have been plenty of changes

BuffYoda21:
Small note, in order for intellesense not to get upset for trying to reference the json files, contributors will need to add `"${workspaceFolder}/build"` to their includePath in `c_cpp_properties.json`. Not sure if there is another better way to do this since that will mean if you haven't built the app yet or run `make clean` intellesense will get upset.

data/item_table.jsonc
viccie211, line 1: I understand why this file is formatted this way because it reads correct in an editor. But it doesn't really work in the Github interface, Which means that PRs would be quite the hassle reviewing it.

source/descriptions.cpp
```diff
@@ -1,2166 +1,47 @@
...
-                                                        "through the wall. A bomb can be used to push Link\n"
-                                                        "fully through the wall to get out of bounds. This\n"
-                                                        "can be used in the fire temple along with hover\n"//
-                                                        "boots to skip hammer in the first room.";         //
+void DescriptionTable_Init() {
+    const char* data = reinterpret_cast<const char*>(description_table_jsonc);
+    std::string text(data, description_table_jsonc_size);
+
+    json table = json::parse(text.begin(), text.end(), nullptr, true, true);
+    for (auto i = table.begin(); i != table.end(); ++i) {
```
viccie211, line 17:
```suggestion
    for (auto i = table.begin(); i < table.end(); ++i) {
```

source/descriptions.cpp
```diff
@@ -1,2166 +1,47 @@
...
+                }
+                joined += entry[j].get<std::string>();
+            }
+            descriptionTable.emplace(key, std::move(joined));
+        } else if (entry.is_string()) {
+            descriptionTable.emplace(key, entry.get<std::string>());
+        }
+    }
+}
+
```
viccie211, line 36: I'm guessing you need to loop through the entries of a JSON table for all the tables. So would it be worth it to generalize that code, so parsing the json will be done in one place?

BuffYoda21: I had considered moving all the json reading to it's own file but decided to at least wait on it for now since only one file is being read right now. Still may change in the future though. Also the parsing is different between each json file since they store entirely different types of data so I don't think there is a clean way to generalize it any further and it would just really be moving all the json reading methods to one place. Though it would make the include statements a bit easier to work with.

data/item_table.jsonc
BuffYoda21: If you view the file it's self rather than the PR diff (since it is all additions anyway) with the three dots in the top right of the diff pane it looks a lot cleaner. Though I can also temporarily remove all these other json files that aren't in use yet since I am not entirely sure about their schema's yet anyway.

BuffYoda21: I was also just trying to mirror the original schema as close as possible

BuffYoda21: I am going to go ahead and remove the unused data files as they are not even up to date with more recent commits. `description_table.jsonc` is the only one that is really ready for review now anyway.

BuffYoda21:
It turns out this change [ed982c3](https://github.com/gamestabled/OoT3D_Randomizer/pull/807/commits/ed982c3e35f168a39d36bd71b64db8c42f591174) actually causes the app to crash on startup because nlohman/json doesn't support `<` `>` operators when iterating. `!=` is the correct way to do it.

PR #808, opened 2026-04-01
Improve address management

HylianFreddy:
This PR aims to improve the build time by removing the need to recompile files.

Patch side changes:
- All addresses for base game functions/variables are now located in the linker script, and all macros casting a number to a function/data type have been replaced with a normal declaration. This allows using the same build output for USA and EUR regions.
- The formatting/structure of the linker script is changed to be more compact.
- The .s files have been moved to a subdirectory `asm`.
- The patches and hooks in the asm code are now defined with macros to avoid repetition.
- The patch Makefile is updated to use the same build directory for both regions.
- Region flags are not passed to the compiler or assembler anymore, only the linker will get them.
- The python script `ld_addr.py` can be used to add new symbols or patches to the linker script.
- The `patch.py` script has been split, moving all app side stuff out of it.
- The `NEWCODE_SIZE` value (used to repermission memory to be executable) will now be calculated by the linker instead of being hardcoded to a high value.
- Some functions have been renamed for clarity and consistency with the OoT decomp repo
- Other minor improvements (removing unused code, typing action functions,...)

App side changes:
- The `PatchSymbols` struct is now defined in a file generated by the script `generate_patch_symbols.py`, so if its values change only a single small translation unit will need to be recompiled. There is also only one struct for both regions now, as the base address for custom code is the same in both, so all the symbols are always the same too.


gymnast86 (approved):

code/src/actors/chest_minigame.c
```diff
@@ -18,12 +17,12 @@ void EnTakaraMan_rInit(Actor* thisx, GlobalContext* globalCtx) {
...
-        sLoserGetItemIds[2] = 0x7C;
-        sLoserGetItemIds[3] = 0x7C;
-        sLoserGetItemIds[4] = 0x7C;
-        sLoserGetItemIds[5] = 0x7C;
+        EnChanger_LoserGetItemIds[0] = 0;
+        EnChanger_LoserGetItemIds[1] = 0x7C;
+        EnChanger_LoserGetItemIds[2] = 0x7C;
+        EnChanger_LoserGetItemIds[3] = 0x7C;
+        EnChanger_LoserGetItemIds[4] = 0x7C;
+        EnChanger_LoserGetItemIds[5] = 0x7C;
```
VictorRemmerswaalAuxilium, line 25: I know you just renamed the variable, but wouldn't this be better done in a loop?

VictorRemmerswaalAuxilium (approved):
It kinda goes over my head. Especially the Python and linker bit, but the code looks way cleaner this way. Also I built and ran the app and it generates a proper usable seed.

code/src/actors/chest_minigame.c
HylianFreddy: Yes, but I'd rather not change code logic in this PR to keep it focused on just the symbol declarations.

PR #810, opened 2026-04-08
Move shared declarations into dedicated headers

HylianFreddy:
Currently the app's `.cpp` files include some headers from the base patch with ugly directives like `#include "../code/src/entrance.h"` that cause a bunch of useless declarations to be visible from the app files and create useless dependencies for Make (the stuff in the `.d` files).

So, I moved all declarations that are used by both the patch and the app into separate headers under the `./shared` directory, including them where required. I opted to name them all with a `s_` prefix for clarity and to avoid include conflicts.
I also placed the few shared definitions in a `shared_data.c` file which will be compiled for both subprojects.

To avoid error squiggles in VS Code, developers will need to update their local `c_cpp_properties.json`, adding something like this to the include path:
`"${workspaceFolder}/shared/**"`


code/src/actors/owl.h
```diff
@@ -5,6 +5,5 @@
 
 void EnOwl_DespawnInit(Actor* owl, GlobalContext* globalCtx);
 void EnOwl_rUpdate(Actor* owl, GlobalContext* globalCtx);
-void EnOwl_FastCutscene(GlobalContext* globalCtx);
```
viccie211, line 8: Was this unused? Otherwise, why remove it?

viccie211 (approved):

code/src/actors/owl.h
HylianFreddy: Yeah it was unused. I probably should've deleted it in the previous PR but I noticed it late 😅

PhlexPlexico (approved):

PR #813, opened 2026-04-12
Add fanfare overrides

HylianFreddy:
This PR adds some overrides for the "item get" fanfare. For trap items the pitch will decrease linearly every frame (thanks Nessy for the suggestion). For Medallions and Spiritual Stones, the fanfare will be the one used during their vanilla cutscene instead of the default one.
I've also renamed some audio related functions and added an enum for Sequence Ids.

https://github.com/user-attachments/assets/a6d1e51c-8255-4283-852d-c20244ede365


code/src/bgm.c
```diff
@@ -26,3 +28,21 @@ u32 SetBGM(u32 original) {
...
+    if (!Bgm_FanfareModEnabled) {
+        return;
+    }
+    if (gUnkSequencePlayerData[SEQ_PLAYER_FANFARE] == NULL) {
+        Bgm_FanfareModEnabled = FALSE;
+        return;
+    }
+
+    // Linearly decrease pitch
+    gUnkSequencePlayerData[SEQ_PLAYER_FANFARE]->freq -= 0.009;
```
viccie211, line 47: Maybe make this a const instead of a magic number? Then it would be easier in the future to change it with a setting or a variable.

viccie211 (approved):

code/src/bgm.c
HylianFreddy: I thought it was fine because of the comment. Also the value is only used here, not in multiple places. Would you prefer something like this? What name would you suggest for the variable?
```
const f32 var = 0.009;
gUnkSequencePlayerData[SEQ_PLAYER_FANFARE]->freq -= var;
```

Roberto-Nessy (approved):

PR #815, opened 2026-04-17
Add cosmetic option for Ganon/dorf's blood colour

Roberto-Nessy:
Options:
- Vanilla green (default)
- N64 1.0 red
- Random
- Custom

### DIFF
--- a/code/src/main.c
+++ b/code/src/main.c
@@ -40,7 +40,7 @@ void Randomizer_Init() {
     IceTrap_Init();
     extDataInit();
     irrstInit();
-    initGanonBloodColors();
+    Effects_Init();
 
     s64 output = 0;
     svcGetSystemInfo(&output, 0x20000, 0);
--- a/code/src/main.c
+++ b/code/src/main.c
@@ -3,6 +3,7 @@
 #include "actor.h"
 #include "input.h"
 #include "models.h"
+#include "effects.h"
 #include "entrance.h"
 #include "settings.h"
 #include "title_screen.h"
@@ -39,6 +40,7 @@ void Randomizer_Init() {
     IceTrap_Init();
     extDataInit();
     irrstInit();
+    initGanonBloodColors();
 
     s64 output = 0;
     svcGetSystemInfo(&output, 0x20000, 0);
--- a/source/settings.cpp
+++ b/source/settings.cpp
@@ -3091,17 +3091,18 @@ static void UpdateCosmetics() {
     }
     // Ganon/dorf Blood
     switch (Settings::GanonBloodColor.Value<u8>()) {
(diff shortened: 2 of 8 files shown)


code/src/main.c
```diff
@@ -40,6 +40,7 @@ void Randomizer_Init() {
     IceTrap_Init();
     extDataInit();
     irrstInit();
+    initGanonBloodColors();
```
HylianFreddy: Could you rename this function to something like `Effects_Init`, so it follows the same pattern of the other ones?

source/settings.cpp
```diff
@@ -1268,6 +1268,7 @@ Option BoomerangTrailDuration     = Option::U8  (2, "Boomerang (Duration)",  tra
...
 Option BombchuTrailOuterColor     = Option::U8  (2, "Bombchu (Outer Color)", weaponTrailOuterOptionNames,   {RANDOM_CHOICE_DESC, RANDOM_COLOR_DESC, CUSTOM_COLOR_DESC, "Select the color for the sides of the\nbombchu trail."},                                              OptionCategory::Cosmetic,    SAME_AS_INNER_TRAIL);
 Option BombchuTrailDuration       = Option::U8  (2, "Bombchu (Duration)",    chuTrailDurationOptionNames,   {"Select the duration for bombchu trails."},                                                                                                                      OptionCategory::Cosmetic,                      2); // Vanilla
+Option GanonBloodColor            = Option::U8  ("Ganon/dorf Blood Color",   {"Vanilla Green", "Original Red", std::string(RANDOM_COLOR_STR), std::string(CUSTOM_COLOR_STR),}, {"Select the color of Ganondorf and Ganon's blood."},                                          OptionCategory::Cosmetic);
```
HylianFreddy:
```suggestion
Option GanonBloodColor            = Option::U8  ("Ganon/dorf Blood Color",   {std::string(RANDOM_COLOR_STR), std::string(CUSTOM_COLOR_STR), "Vanilla Green", "Original Red"}, {RANDOM_COLOR_DESC, CUSTOM_COLOR_DESC, "Select the color of Ganondorf and Ganon's blood."},     OptionCategory::Cosmetic,     BLOODCOLOR_VANILLA);
```
To keep the description for random/custom color consistent with the other settings you'd have to rearrange the options to have random/custom color at the beginning.

HylianFreddy (changes requested):
Cool setting!

source/settings.cpp
Roberto-Nessy: Added extra defaults (from the weapon trails list) so it'd make sense to include the random choice option too and use the `ChooseFinalColor` helper

HylianFreddy (approved):

PR #817, opened 2026-04-30
Fixes spoiler item checks for Rupeesanity items when Rupeesanity is off.

viccie211:
Sets the spoiler checks of freestanding rupees to REVEALTYPE_ALWAYS if rupeesanity is off.

Clashes with #818
We should determine which one is preferrable

### DIFF
--- a/source/spoiler_log.cpp
+++ b/source/spoiler_log.cpp
@@ -278,7 +278,8 @@ void WriteIngameSpoilerLog() {
             splrDatLoc->ItemLocations[spoilerItemIndex].RevealType  = REVEALTYPE_ALWAYS;
         }
         else if(loc->IsCategory(Category::cFreestandingRupee) && Settings::ShuffleRupees.Value<bool>() == false){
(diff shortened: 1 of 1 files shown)


source/spoiler_log.cpp
```diff
@@ -277,6 +277,11 @@ void WriteIngameSpoilerLog() {
             splrDatLoc->ItemLocations[spoilerItemIndex].CollectType = COLLECTTYPE_REPEATABLE;
             splrDatLoc->ItemLocations[spoilerItemIndex].RevealType  = REVEALTYPE_ALWAYS;
         }
+        else if(loc->IsCategory(Category::cFreestandingRupee) && Settings::ShuffleRupees.Value<bool>() == false){
```
viccie211: `.Is(OFF)` crashed the rando app. This does work

HylianFreddy: You can use Option objects like boolean values. So you can check that the setting is disabled with:
```!Settings::ShuffleRupees```

PR #818, opened 2026-04-30
Skipped the free standing rupees in the ingame spoiler log when rupeesanity is disabled

viccie211:
Removes the ingame spoiler log locations for the freestanding rupees when rupeesanity is disabled.

Clashes with #817 
We should determine which one is preferrable.

### DIFF
--- a/source/spoiler_log.cpp
+++ b/source/spoiler_log.cpp
@@ -192,7 +192,7 @@ void WriteIngameSpoilerLog() {
                  (Settings::GerudoFortress.Is(GERUDOFORTRESS_FAST) && loc->IsCategory(Category::cVanillaGFSmallKey) &&
                   loc->GetHintKey() != GF_NORTH_F1_CARPENTER)) {
             continue;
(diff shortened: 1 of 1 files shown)


source/spoiler_log.cpp
```diff
@@ -192,6 +192,8 @@ void WriteIngameSpoilerLog() {
                  (Settings::GerudoFortress.Is(GERUDOFORTRESS_FAST) && loc->IsCategory(Category::cVanillaGFSmallKey) &&
                   loc->GetHintKey() != GF_NORTH_F1_CARPENTER)) {
             continue;
+        } else if (loc->IsCategory(Category::cFreestandingRupee) && !Settings::ShuffleRupees) {
```
HylianFreddy: Could you add a comment like this to keep the formatting consistent?

```suggestion
        }
        // Freestanding Rupees
        else if (loc->IsCategory(Category::cFreestandingRupee) && !Settings::ShuffleRupees) {
```

HylianFreddy (approved):

PR #820, opened 2026-05-07
Skip overrides for vanilla items

HylianFreddy:
This PR removes the item overrides from the patch for locations that are set to be vanilla (due to vanilla logic or disabled shuffle settings). This fixes freestanding rupees appearing as items even when they're not shuffled.

In the `ItemLocation` class I refactored the `hidden` field, previously used only to hide spoiler log entries, renaming it to `overridden` to indicate if that location will have an override placed in the patch. Placing a random item will set it to true, and placing the vanilla item will set it to false.

I also updated the business scrub code and removed the "hack" for seeds/arrows scrubs. I tested to make sure the scrubs still work correctly, checking both scrub types (respawning and non-respawning ones) with the setting enabled, disabled, and with vanilla logic.

For `Delayed` and `TempleReward` item locations I chose to keep the overrides even when vanilla, because those items are always given by the rando code. I added an enum for delayed overrides to make them clearer.


Kewlan (dismissed):

shared/s_item_override.h
Roberto-Nessy, line 1: Minor/inconsequential nitpick, but I'd prefer the prefix be `DLYOVR_` rather than `DELOVR_` simply because I'm used to "DEL" meaning delete

HylianFreddy: Done, thanks for the suggestion!

Roberto-Nessy (approved):

PR #824, opened 2026-05-18
Add setting to keep extra shields

HylianFreddy:
This adds a setting to allow keeping extra shields found in randomized locations even when the player already owns that type of shield. This option will be enabled by default. There is also an extra option to allow buying extra shields from shops, which normally refuse to sell if you have the shield.


viccie211 (approved):

Kewlan (approved):

ShawKai91:
What would this do in practice?

HylianFreddy:
> What would this do in practice?

This is the setting description:

```
Allow keeping more than 1 Deku and Hylian shield
in the inventory, so if you lose one you can then
re-equip it immediately. The shield count will be
displayed next to the item name.
You can choose if extra shields should only be
obtainable from randomized items or if they should
also be repeatedly buyable from shops.
```

PR #825, opened 2026-05-18
Small improvements for enemy features

HylianFreddy:
This adds the following changes for Enemy Soul Shuffle and Enemy Randomizer:
- doors will no longer lock behind Link when the room contains soulless enemies, so that the player can simply go back out instead of being forced to savewarp or teleport away;
- randomized Dark Link will no longer be stuck in a falling animation when spawning in certain places, and will instead appear in the normal idle standing pose;
- Dead Hand will now detect explosions and come out of the ground even when soulless, similar to how Armos enemies can detect projectiles and wake up.

### DIFF
--- a/code/src/actors/door.c
+++ b/code/src/actors/door.c
@@ -2,6 +2,7 @@
 #include "models.h"
 #include "settings.h"
 #include "enemizer.h"
+#include "enemy_souls.h"
 #include "multiplayer.h"
 
 // Certain doors can cause a crash depending on a freestanding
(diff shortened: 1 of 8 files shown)


code/src/actors/door.c
```diff
@@ -96,32 +99,45 @@ void DoorShutter_rUpdate(Actor* thisx, GlobalContext* globalCtx) {
...
     }
-    this->action_fn = DoorShutter_SlidingDoor_Idle;
+    this->actionFunc = DoorShutter_Idle;
+}
+
+u8 DoorShutter_CheckSoullessEnemies(DoorShutter* this) {
+    if (gSettingsContext.shuffleEnemySouls == SHUFFLEENEMYSOULS_ALL &&
+        (this->actionFunc == DoorShutter_Close || this->actionFunc == DoorShutter_Open)) {
+        Actor* enemy = gGlobalContext->actorCtx.actorList[ACTORTYPE_ENEMY].first;
+        for (; enemy != NULL; enemy = enemy->next) {
```
viccie211, line 133: There is no possibility here that this loop will go on forever or that the loop will go read out of bounds?

viccie211 (approved):

code/src/actors/door.c
HylianFreddy: No, because actors are stored in a linked list where the last actor has NULL as the next pointer, so this loop will go through the list and stop at the end. This is also the way the base game parses all actors in a category, here's an example from the OoT decomp: https://github.com/zeldaret/oot/blob/4834ec2b6913a109bca9a2249715ab08f0f6ad42/src/code/z_actor.c#L4667

Kewlan (approved):

PR #828, opened 2026-05-29
Logic updates

jcallen:
Add a couple of logic updates to prevent impossible seeds (without glitches)

### DIFF
--- a/source/location_access/locacc_shadow_temple.cpp
+++ b/source/location_access/locacc_shadow_temple.cpp
@@ -161,7 +161,7 @@ void AreaTable_Init_ShadowTemple() {
                 // Exits
                 Entrance(SHADOW_TEMPLE_WIND_TUNNEL, { [] {
                              return (LogicLensShadowBack || CanUse(LENS_OF_TRUTH)) && Hookshot &&
-                                    SmallKeys(SHADOW_TEMPLE, 3, 4);
+                                    SmallKeys(SHADOW_TEMPLE, 3, 4) && (CanDefeatEnemies(7, 0, 11) || CanUse(LONGSHOT));
                          } }),
             });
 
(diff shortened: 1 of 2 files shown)


source/location_access/locacc_shadow_temple.cpp
```diff
@@ -161,7 +161,7 @@ void AreaTable_Init_ShadowTemple() {
                 // Exits
                 Entrance(SHADOW_TEMPLE_WIND_TUNNEL, { [] {
                              return (LogicLensShadowBack || CanUse(LENS_OF_TRUTH)) && Hookshot &&
-                                    SmallKeys(SHADOW_TEMPLE, 3, 4);
+                                    SmallKeys(SHADOW_TEMPLE, 3, 4) && (CanDefeatEnemies(7, 0, 11) || CanUse(LONGSHOT));
```
HylianFreddy, line 164: This can actually be done with the base hookshot even without spawning the chest. It's different from the original OoT because the hook travels a tiny bit further in this version. To do it you need to stand where the chest would be and aim at the target, the reticle will not appear but the hook will connect anyway.

You could add a logic trick setting and make this entrance require either the longshot or the trick. I don't think the check on the enemies should be there, because even jumping on the chest should be a trick anyway.

write_version.sh
HylianFreddy, line 1: Was this empty file committed accidentally?

HylianFreddy (changes requested):

write_version.sh
jcallen: That isn't an empty file, nor is it an empty change -- it's a file permissions update (setting the script to be executable). Note the "100644 -> 100755" above the diff.

HylianFreddy: Oh right, sorry I'm not used to look at the file mode in git. What does this change in practice? The build script still needs to run `chmod +x linux_build_rando.sh write_version.sh` before running those scripts, right?

jcallen: If the files are marked as executable in the repository, you shouldn't need to have the build script change the permissions (unless GitHub Actions does something unusual).  As I am developing on Linux directly, once the permissions are updated in the repository, no further permissions changes are required to get the build to work in a local container.

jcallen: Permissions changes split into separate PR (#829).

source/location_access/locacc_shadow_temple.cpp
HylianFreddy: Hi, are you still interested in implementing this logic trick? If you prefer I can make it myself and then merge this PR. But if you just need more time that's fine too, just let me know :)

PR #837, opened 2026-06-17
Add Big Poe Shuffle

HylianFreddy:
This new setting will make the 10 Big Poes in Hyrule Field drop randomized items. Their spirits will be shuffled into the item pool and won't require a bottle to carry, instead they will immediately give 100 points on the Poe Collector's card.

Todo:
- [ ] Translations for new messages


code/object_and_gi_usage.txt
```diff
@@ -13,6 +13,7 @@ The following custom objectIds are currently being used:
 127: Ganon Small Key
 128: Boss Keys
 131: Rupoor
+131: Unbottled Big Poe
```
jcallen:
```suggestion
132: Unbottled Big Poe
```

Based on the changes to `enum ObjectId` in `objects.h`

HylianFreddy: Oops, thanks for catching that!

Kewlan (dismissed):

Kewlan (approved):

PR #842, opened 2026-07-03
Fixed GF key logic in fast setting

Roberto-Nessy:
Reverted skipped guards' items to recovery hearts as they were before #820
Did not revert the items for the open setting as is inconsequential


HylianFreddy (approved):
I don't know how I missed that the items placed here were hearts instead of keys, oops :sweat_smile:

viccie211 (approved):

PR #846, opened 2026-07-17
Add sub-option for Playthrough Hints Specificity

HylianFreddy:
This adds a small sub-setting to the Playthrough Hints distribution, to make them reveal the exact location of the item instead of just the general region. It can be useful when shuffling a lot of things, or just to have a faster game.


source/hints.cpp
```diff
@@ -242,10 +242,15 @@ static void CreateAreaLocationHint(LocationKey hintedLocation) {
...
-
-    Text finalHint = prefix + "#" + itemHintText + "# " + foundAt + " #" + areaHintText + "#.";
+    Text finalHint;
+    if (HintSpecificity.Is(HINTSPECIFICITY_GENERAL)) {
+        Text foundAt      = Hint(CAN_BE_FOUND_AT).GetText();
+        Text areaHintText = Hint(GetLocationRegionHintKey(hintedLocation)).GetText();
+        finalHint         = prefix + "#" + itemHintText + "# " + foundAt + " #" + areaHintText + "#.";
+    } else { // HINTSPECIFICITY_EXACT
+        Text locHintText = Location(hintedLocation)->GetHint().GetText();
+        finalHint        = prefix + locHintText + " #" + itemHintText + "#.";
```
Roberto-Nessy, line 252: I feel like there should be some preposition between the location and item? Or does that create hints that are either too long or don't make sense in other languages?

HylianFreddy: No this is just the regular hint structure, used in `CreateLocationHint` for example: https://github.com/gamestabled/OoT3D_Randomizer/blob/main/source/hints.cpp#L214
Prepositions aren't needed because location hints already contain them, like "#pillars of wood# in the Deku Tree lead to" for DEKU_TREE_COMPASS_CHEST.

Roberto-Nessy: Oh cool, looks good to me then

Roberto-Nessy (approved):

PR #849, opened 2026-08-09
Added a time multiplier for the final frog song to make it more lenient

viccie211:


viccie211:
Oh I seem to have bungled it up. Please wait with reviewing.

viccie211:
Fixed it, it's ready for review again

code/src/asm/hooks.s
```diff
@@ -374,6 +374,15 @@ HOOK KingZoraCheckMovedFlag
...
     bx lr
 
+HOOK EnFrSetupFrogSongTimerMultiplier
+    push {r0,r1,r3-r12,lr}
+    mov r0,r2
+    bl FrogSongTimerMultiplier
+    mov r2,r0
+    pop {r0,r1,r3-r12,lr}
+    strh r2,[r1,#0x42]
+    b 0x381208
```
viccie211: Will this branch be linked to the correct offset for the EU version? Or do I need to do something else with it?

shared/s_settings.h
```diff
@@ -813,5 +813,6 @@ typedef struct SettingsContext {
     u8 startingOcarinaButtons;
 
     u8 startingTokens;
+    u8 frogSongTimerMultiplier;
```
HylianFreddy: For consistency, I would move this up, after `u8 randomSongNotes`.

source/descriptions.cpp
```diff
@@ -1458,6 +1458,11 @@ string_view gsLocGuaranteeNewDesc     = "Excludes the original location from the
 string_view randomSongNotesDesc       = "Randomize the notes for each ocarina song.\n"     //
                                         "Regular songs will be 3 notes repeated twice.\n"  //
                                         "Warp songs will be between 5 and 8 notes.";       //
+/*------------------------------                                                           //
+|      FROG SONG TIMER         |                                                           //
+------------------------------*/                                                           //
+string_view frogSongTimeMultDesc      = "Multiplier for the time you have to play each\n"  //
+                                        "note in the final frog song.";                  //
```
HylianFreddy: Formatting
```suggestion
                                        "note in the final frog song.";                    //
```

code/src/actors/frog.c
```diff
@@ -13,3 +13,13 @@ void EnFr_rSetReward(EnFr* frog) {
...
+
+s16 FrogSongTimerMultiplier(s16 initialFrogTimer) {
+    return initialFrogTimer * (gSettingsContext.frogSongTimerMultiplier + 1);
+}
+
+void FrogSongTimer_Init(void) {
+    for (u8 i = 0; i < ARRAY_SIZE(sTimerFrogSong); i++) {
+        sTimerFrogSong[i] = FrogSongTimerMultiplier(sTimerFrogSong[i]);
+    }
+}
```
HylianFreddy: Formatting
```suggestion
}

```

code/src/main.c
```diff
@@ -24,8 +24,10 @@
 #include "bgm.h"
 #include "business_scrubs.h"
 #include "alert.h"
+#include "frog.h"
 
 #include "z3D/z3D.h"
+#include "z3D/actors/z_en_fr.h"
```
HylianFreddy: This is unused
```suggestion
```

source/settings.cpp
```diff
@@ -1849,7 +1851,8 @@ SettingsContext FillContext() {
     ctx.startingOcarinaButtons |= StartingOcarinaButtonY.Value<u8>() << 3;
     ctx.startingOcarinaButtons |= StartingOcarinaButtonA.Value<u8>() << 4;
 
-    ctx.startingTokens = StartingSkulltulaToken.Value<u8>();
+    ctx.startingTokens          = StartingSkulltulaToken.Value<u8>();
+    ctx.frogSongTimerMultiplier = FrogSongTimeMultiplier.Value<u8>();
```
HylianFreddy: Like for the declaration, I would move this after `ctx.randomSongNotes`

source/settings.cpp
```diff
@@ -487,6 +487,7 @@ Option FreeCamera          = Option::Bool("Free Camera",            {"Off", "On"
 Option RandomGsLocations   = Option::Bool("Random GS Locations",    {"Off", "On"},                                                          {randomGsLocationsDesc});
 Option GsLocGuaranteeNew   = Option::Bool(2, "Guarantee New",       {"Off", "On"},                                                          {gsLocGuaranteeNewDesc});
 Option RandomSongNotes     = Option::Bool("Random Ocarina Melodies",{"Off", "On"},                                                          {randomSongNotesDesc});
+Option FrogSongTimeMultiplier = Option::U8("Frogsong Time Multiply",{"1x","2x","3x","4x"},                                                  {frogSongTimeMultDesc});
```
HylianFreddy: I think a simpler name like "Frog Song Timer" would look better. Also, how about shortening the setting variables names to `frogSongTimerMult`? That way they would fit the indentation.

HylianFreddy (changes requested):
This looks good to me, I just have some code style suggestions.

code/src/asm/hooks.s
HylianFreddy: This is fine, addresses outside the range 0x419E18-0x4A5AFF are the same in both versions. And there are a lot of hard-coded addresses in hooks.s still, so this won't be an outlier

Roberto-Nessy: Personally I'd recommend using a `bl` in the hook and `bx lr` here, there are times where hard-coded addresses are the only feasible option but I'd prefer to keep them to a minimum.

This isn't always available as an option, but because the function we're patching has already stored the link register on the stack and it won't be used again until the next `bl` where it'll be overwritten a couple lines later it's safe to do in this instance.

viccie211: Oh I will change that then!

HylianFreddy (approved):

Roberto-Nessy (approved):

PR #850, opened 2026-08-11
Improvements for traps, add new curse types

HylianFreddy:
- Add 2 new curses:
  - Invisible Actors (to parallel the Invisible Terrain curse);
  - Missing Navi, preventing not only Navi interactions, but also targeting enemies and objects.
- Allow curses to override active curse, instead of making every curse trap become a bomb/scale trap when a curse is already active;
- Reset Link's action when opening ice trap chests, instead of forcing a falling state to interrupt the get item animation;
- Don't draw slash effect during sword curse, to give a visual indicator of the effect
  - While implementing this, I've also thrown in an extra option to the Trail Duration setting, called "Space Tearing", which keeps the trail effects visible for around 8 seconds.
- Improve receiving multiple traps in a row from pending overrides (like randomized rupees or hearts), because previously they would often reuse the same type due to sharing the same hash value as the last trap;
- General improvements to trap code.


code/src/effects.c
```diff
@@ -14,6 +43,25 @@ void Effects_Init(void) {
     gGanondorfBlood = gGanonBlood = (BloodCol){ col, col };
 }
 
+void Effects_UpdateSwordTrailDuration(void) {
+    u8 duration;
+    if (IceTrap_ActiveCurse == ICETRAP_CURSE_SWORD) {
+        // Don't draw slash effect during sword curse.
+        duration = 0;
```
viccie211: Maybe better to use the  `sSwordTrailDurationValues[TRAILDURATION_DISABLED]` then too? I know it also results in 0 but still it reads better?

HylianFreddy: Good idea, done!

viccie211 (approved):

Roberto-Nessy (approved):

